PR #1, opened 2016-07-01
Improve horizontal swipe detection

satya164:
cc @jsierles

### DIFF
--- a/src/TabViewPanResponder.js
+++ b/src/TabViewPanResponder.js
@@ -1,5 +1,6 @@
 /* @flow */
 
+import { Platform } from 'react-native';
 import type { GestureEvent, GestureState } from './PanResponderTypes';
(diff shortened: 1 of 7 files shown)


src/TabViewPanResponder.js
```diff
@@ -8,7 +9,7 @@ type Props = SceneRendererProps & {
 }
 
 const POSITION_THRESHOLD = 120;
-const VELOCITY_THRESHOLD = 0.000001;
+const VELOCITY_THRESHOLD = Platform.OS === 'android' ? 0.0000005 : 0.5; // on Android, velocity is way lower, perhaps due to timestamp being in nanosecond
```
satya164: There's a open PR which fixes this, we can remove this condition when it goes to stable - https://github.com/facebook/react-native/pull/8199

PR #4, opened 2016-07-04
refactor: rename TabBar class

damusnet:


knowbody:
thanks, good catch! 👍

PR #6, opened 2016-07-04
Revert "feat: pass position to scene for label animation"

satya164:
Reverts react-native-community/react-native-tab-view#5

### DIFF
--- a/src/TabBar.js
+++ b/src/TabBar.js
@@ -85,7 +85,6 @@ export default class TabBar extends Component<DefaultProps, Props, void> {
             route,
             focused,
             index: i,
-            position,
           };
           const icon = this.props.renderIcon ? this.props.renderIcon(scene) : null;
           const label = this.props.renderLabel ? this.props.renderLabel(scene) : null;


satya164:
@damusnet Hey, sorry for reverting the PR. I've couple of points regarding this,

We don't really need to pass position to `renderLabel`, you can always use a partially applied function, which is a lot cleaner with ES2015 ❤️

```js
renderLabel = position => scene => {
  // do stuff
};
```

We pass position, we might pass layout, and other stuff as well. But I would prefer to avoid that since it's not really necessary.

Right now `renderLabel`, `renderIcon`, `renderScene` etc have the same signature (`scene: { route, focused, index }`), and `renderHeader`, `renderFooter`, `renderIndicator` get all the props, which makes sense, since scene specific items get the scene as a prop, and other top level items get all the props. It also makes it easier to learn.

Thanks for the PR though. Much appreciated 😍

damusnet:
The PR was only meant as a suggestion, so I totally get it.
Thanks for taking the time to explain your reasoning!

PR #7, opened 2016-07-06
allow opacity prop to be an object

knowbody:
i.e. that's what CoverflowExample does


satya164:
It should actually be `Animated.View.PropTypes.style`, otherwise if you add other animated values, you'll get more warnings.

But this creates another issue. We pass the animation value for index as a prop named `position`. But if we add the propType for style, it tries to validate `position` also and generates a false warning that `position` should be passed as style rather than a prop directly.

Closing this for now until we find a way to suppress the position related warning.

PR #12, opened 2016-08-08
fix: improve TouchableItem press animation

### LINKED ISSUE #11: Improve TouchableItem press animation
If you touch quickly on the tab item button, the ripple effect won't trigger, so what I suggest is that you put ```delayPressIn: 0``` in defaultProps to give a nice feedback effect when the button is pressed, no matter how fast you touch the screen.

I think this should be the default in React Native, maybe some people don't want this, so it would be nice if there's a way to pass props to TouchableItem through TabBarTop and TabBar.

Great component btw :+1:

fmatoss:
Fixes #11

### DIFF
--- a/src/TouchableItem.js
+++ b/src/TouchableItem.js
@@ -18,6 +18,7 @@ type Props = {
 }
 
 type DefaultProps = {
+  delayPressIn: number;
   pressColor: string;
 }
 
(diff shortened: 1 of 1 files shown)


satya164:
Would be awesome if you could also add the propType.

fmatoss:
Forgot about it.

satya164:
Merged. Thanks :D

PR #14, opened 2016-08-16
added labelStyle prop to TabBarTop component

knowbody:


example/src/TopBarTextExample.js
```diff
@@ -48,6 +48,7 @@ export default class TopBarTextExample extends Component {
         pressColor='rgba(0, 0, 0, .2)'
         indicatorStyle={styles.indicator}
         style={styles.tabbar}
+        labelStyle={{ color: '#fbdc08', fontWeight: '500' }}
```
satya164: Move this to the `StyleSheet.create` call. Also probably just make it `{ color: '#fff', fontWeight: '500' }` since I want to show that the default styles are based on material design.

src/TabBarTop.js
```diff
@@ -27,30 +27,28 @@ const styles = StyleSheet.create({
...
-}
-
 type Props = SceneRendererProps & {
   renderLabel: (scene: Scene) => React.Element<any>;
   indicatorStyle?: any;
+  labelStyle?: any;
 }
 
-export default class TabBarTop extends Component<DefaultProps, Props, void> {
+export default class TabBarTop extends Component<Props, void> {
```
satya164: this should be `export default class TabBarTop extends Component<void, Props, void> {`

PR #27, opened 2016-09-19
fix: remove `shallowCompare` in `shouldComponentUpdate`

satya164:
Having `shouldComponentUpdate` implemented inside the library has proved to be more confusing than helpful. We still need to ensure minimal updates to have performant animations and gestures. So we recommend users to implement `shouldComponentUpdate` and have it in the README. We do minimal `setState` inside the library, so there should be no unnecessary rendering caused by the library code, but users code.

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,6 +27,7 @@ npm install --save react-native-tab-view react-addons-shallow-compare
 
 ```js
 import React, { Component } from 'react';
+import shallowCompare from 'react-addons-shallow-compare';
 import { View, StyleSheet } from 'react-native';
 import { TabViewAnimated, TabViewPage, TabBarTop } from 'react-native-tab-view';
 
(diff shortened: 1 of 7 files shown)


knowbody:
It was confusing, but I'm not sure if that's a solution... Now I either have unoptimised and slow tab view. Or find out that I actually need `shallowCompare` and will need to install it anyway. I think the way it is now is good, but again, I don't think there is a silver bullet for it.

I think it is worth mentioning somewhere how `shallowCompare` works, as probably not everyone uses/used it. Maybe redirecting to the React docs: https://facebook.github.io/react/docs/shallow-compare.html

satya164:
Honestly I'm not sure if this is the correct fix either. Me and @skevy discussed about this a bit yesterday and decided to add a `shouldUpdate` prop to allow choosing between both behaviours.

PR #28, opened 2016-09-20
feat: add a shouldOptimizeUpdates prop to customize shouldComponentUpdate behaviour

satya164:
I removed `shouldComponentUpdate` from all components other than `TabViewAnimated` and `TabViewTransitioner` since they are rendered by `TabViewTransitioner`, hence the extra `shouldComponentUpdate` is not necessary.

cc @skevy @knowbody


knowbody:
LGTM!

PR #41, opened 2016-10-18
Added a 'deadzone' before PanResponder capture.

julienvincent:
After playing with this library on a few different devices (iOS and Android) it became apparent that the capture point for horizontal scrolling is too sensitive.

This is even more noticeable on phones like the Galaxy S7 and S7 Edge where it became almost impossible to tap on buttons as the swipe capture would take over onTapStart.

This pull adds a `-20 => 20`pt deadzone before capture begins - in other words, when dragging the motion needs to traverse more than 20pt before the PanResponder becomes responder. Might want to make this configurable via component props - although probably not necessary.

For normal swiping the deadzone isn't noticeable while allowing for more natural UI interaction.

### DIFF
--- a/src/TabViewPanResponder.js
+++ b/src/TabViewPanResponder.js
@@ -10,6 +10,7 @@ type Props = SceneRendererProps & {
 }
 
 const DEAD_ZONE = 20;
+
 function forHorizontal(props: Props) {
   let { swipeVelocityThreshold } = props;
 
--- a/src/TabViewPanResponder.js
+++ b/src/TabViewPanResponder.js
@@ -48,8 +48,8 @@ function forHorizontal(props: Props) {
     const { routes, index } = props.navigationState;
     return (
       isMovingHorzontally(evt, gestureState) && (
-        (gestureState.dx >= 30 && index >= 0) ||
-        (gestureState.dx <= -30 && index <= routes.length - 1)
+        (gestureState.dx >= 20 && index >= 0) ||
+        (gestureState.dx <= -20 && index <= routes.length - 1)
     ));
   }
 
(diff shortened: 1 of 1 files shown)


satya164:
Thank you. This looks reasonable. Can you extract it out to a constant declared at the top? Looks good otherwise.

julienvincent:
Done 👍

src/TabViewPanResponder.js
```diff
@@ -20,6 +20,7 @@ function forHorizontal(props: Props) {
 
   let lastValue = null;
   let isMoving = null;
+  let deadzone = 20;
```
satya164: nit: it's a constant, so can you declare it above the function as `const DEAD_ZONE = 20`? Sorry for nitpicking.

julienvincent: Ah I see :)

src/TabViewPanResponder.js
```diff
@@ -46,10 +46,11 @@ function forHorizontal(props: Props) {
 
   function canMoveScreen(evt: GestureEvent, gestureState: GestureState) {
     const { routes, index } = props.navigationState;
+    const DEAD_ZONE = 20;
```
satya164: I mean declare before `function forHorizontal(props: Props) {`

julienvincent:
:)

satya164:
whoa, it has conflicts now!

julienvincent:
Yea apologies its due to the fact that I let this leak into the swipe PR

julienvincent:
I'll sort out the conflict quick.

julienvincent:
Could you make a release once this is merged? I would like to use it :)

satya164:
@julienvincent sure thing. will do in an hour.

Issue #38, opened 2016-10-16
Unable to 'Cancel' swipes properly


julienvincent:
Thanks for this very well written library. Very performant with a highly reusable API. One issue I did find with it is that its hard to cancel a swipe.

If you start a swipe in a direction, lets say `to the right`, and then attempt to cancel by reversing the swipe `to the left` - the left scene will still open (unless your swipe ends with no velocity and is past the threshold)

julienvincent:
There should probably be some logic detecting if the current motion is in the opposite direction to the initial direction. I will take a look at the source and submit a PR when I get some time, but thought I'd open an issue incase someone wants to get to it sooner.

satya164:
Thank you :)
Can you give a video of what's happening?

![tabs](https://cloud.githubusercontent.com/assets/1174278/19416937/1f591bac-93bc-11e6-9492-0b76ce9619da.gif)

julienvincent:
Sure - how do you make these gif vids? :P

satya164:
@julienvincent this - https://github.com/jclem/gifify

julienvincent:
![swipe mov](https://cloud.githubusercontent.com/assets/8571121/19417028/eebfec9c-93a1-11e6-8e5e-13190ee93583.gif)

julienvincent:
From looking at the source I think the [following line](https://github.com/react-native-community/react-native-tab-view/blob/master/src/TabViewPanResponder.js#L76) is what needs to change. This should contain some logic like `isMoving && !isMovingOppositeDirection`.

Will confirm once I get the example running locally

satya164:
@julienvincent Thanks a lot. We'll probably need to keep track of the latest direction like in `respondToGesture` and then react accordingly.

julienvincent:
@satya164 Agreed 👍

PR #42, opened 2016-10-18
Better handing of a canceled swipe.

julienvincent:
As discussed in #38, this PR adds the necessary logic to resolve that issue.

### DIFF
--- a/src/TabViewPage.js
+++ b/src/TabViewPage.js
@@ -23,9 +23,8 @@ export default class TabViewPage extends Component<void, Props, void> {
       return;
     }
 
-    /* eslint-disable no-console */
+    // eslint-disable-next-line no-console
     console.warn('`<TabViewPage />` is deprecated. Pass the `renderScene` prop to `<TabViewAnimated />` instead.');
-    /* eslint-enable no-console */
     warned = true;
   }
 
--- a/src/TabViewPage.js
+++ b/src/TabViewPage.js
@@ -23,7 +23,9 @@ export default class TabViewPage extends Component<void, Props, void> {
       return;
     }
 
+    /* eslint-disable no-console */
     console.warn('`<TabViewPage />` is deprecated. Pass the `renderScene` prop to `<TabViewAnimated />` instead.');
+    /* eslint-enable no-console */
     warned = true;
   }
 
--- a/src/TabViewPanResponder.js
+++ b/src/TabViewPanResponder.js
@@ -48,8 +48,8 @@ function forHorizontal(props: Props) {
     const { routes, index } = props.navigationState;
     return (
       isMovingHorzontally(evt, gestureState) && (
-        (gestureState.dx >= 30 && index >= 0) ||
-        (gestureState.dx <= -30 && index <= routes.length - 1)
+        (gestureState.dx >= 20 && index >= 0) ||
+        (gestureState.dx <= -20 && index <= routes.length - 1)
     ));
   }
 
--- a/src/TabViewPanResponder.js
+++ b/src/TabViewPanResponder.js
@@ -20,7 +20,7 @@ function forHorizontal(props: Props) {
 
   let lastValue = null;
   let isMoving = null;
-  let startingDirection = null;
+  let startingDirection = 0;
 
   function isIndexInRange(index: number) {
     const { routes } = props.navigationState;
--- a/src/TabViewPanResponder.js
+++ b/src/TabViewPanResponder.js
@@ -36,9 +36,9 @@ function forHorizontal(props: Props) {
 
   function isReverseDirection(gestureState: GestureState) {
     if (startingDirection > 0) {
-      return gestureState.vx < 0
+      return gestureState.vx < 0;
     } else {
-      return gestureState.vx > 0
+      return gestureState.vx > 0;
     }
   }
 
@@ -104,7 +104,7 @@ function forHorizontal(props: Props) {
       if (canMove) {
         startingDirection = gestureState.dx;
       }
-      return canMove
+      return canMove;
     },
     onPanResponderGrant: (evt: GestureEvent, gestureState: GestureState) => {
       startGesture(evt, gestureState);
--- a/src/TabViewPanResponder.js
+++ b/src/TabViewPanResponder.js
@@ -48,8 +48,8 @@ function forHorizontal(props: Props) {
     const { routes, index } = props.navigationState;
     return (
       isMovingHorzontally(evt, gestureState) && (
-        (gestureState.dx >= 0 && index >= 0) ||
-        (gestureState.dx <= 0 && index <= routes.length - 1)
+        (gestureState.dx >= 30 && index >= 0) ||
+        (gestureState.dx <= -30 && index <= routes.length - 1)
     ));
   }
 
--- a/src/TabViewPanResponder.js
+++ b/src/TabViewPanResponder.js
@@ -20,6 +20,7 @@ function forHorizontal(props: Props) {
 
   let lastValue = null;
   let isMoving = null;
+  let startingDirection = null;
 
   function isIndexInRange(index: number) {
     const { routes } = props.navigationState;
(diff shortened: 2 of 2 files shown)


julienvincent:
Oops - I accidentally let another PR leak into this proposal. I will remove the deadzone commits.

src/TabViewPanResponder.js
```diff
@@ -94,12 +94,8 @@ function forHorizontal(props: Props) {
...
 
   return {
-    onStartShouldSetPanResponder: (evt: GestureEvent, gestureState: GestureState) => {
-      return canMoveScreen(evt, gestureState);
-    },
-    onStartShouldSetPanResponderCapture: (evt: GestureEvent, gestureState: GestureState) => {
-      return canMoveScreen(evt, gestureState);
-    },
+    onStartShouldSetPanResponder: () => false,
+    onStartShouldSetPanResponderCapture: () => false,
```
julienvincent: I changed this to return false as sometimes these callouts will fire before the onMove capture and startingDirection does not get set. 

Additionally there shouldn't be any need for these to ever return true (unless they are there for some edge case reason I have not noticed).

src/TabViewPage.js
```diff
@@ -23,7 +23,9 @@ export default class TabViewPage extends Component<void, Props, void> {
       return;
     }
 
+    /* eslint-disable no-console */
     console.warn('`<TabViewPage />` is deprecated. Pass the `renderScene` prop to `<TabViewAnimated />` instead.');
```
satya164: can just use `// eslint-disable-next-line no-console`

src/TabViewPanResponder.js
```diff
@@ -20,6 +20,7 @@ function forHorizontal(props: Props) {
 
   let lastValue = null;
   let isMoving = null;
+  let startingDirection = 0;
```
satya164: nit: change variable name to `startDirection`

src/TabViewPanResponder.js
satya164: @julienvincent Yeah, it looks fine.

src/TabViewPanResponder.js
```diff
@@ -85,17 +94,17 @@ function forHorizontal(props: Props) {
...
-    onStartShouldSetPanResponderCapture: (evt: GestureEvent, gestureState: GestureState) => {
-      return canMoveScreen(evt, gestureState);
-    },
+    onStartShouldSetPanResponder: () => false,
+    onStartShouldSetPanResponderCapture: () => false,
     onMoveShouldSetPanResponder: (evt: GestureEvent, gestureState: GestureState) => {
       return canMoveScreen(evt, gestureState);
     },
     onMoveShouldSetPanResponderCapture: (evt: GestureEvent, gestureState: GestureState) => {
-      return canMoveScreen(evt, gestureState);
```
satya164: Maybe just extract it out to a function like other handlers are

satya164:
Thanks a lot for the PR. I left you minor comments. But looks great!

julienvincent:
Done 👍  Let me know if it needs anything else!

src/TabViewPanResponder.js
```diff
@@ -55,11 +55,14 @@ function forHorizontal(props: Props) {
...
+    const canMove = isMovingHorzontally(evt, gestureState) && (
         (gestureState.dx >= 0 && index >= 0) ||
         (gestureState.dx <= -0 && index <= routes.length - 1)
-    ));
+      );
+    if (canMove) {
+      startDirection = gestureState.dx;
+    }
+    return canMove;
   }
```
julienvincent: Moved the startingDirection capture to this function instead of making a separate one.

satya164:
Thanks. Merged.

julienvincent:
Thanks!

satya164:
I noticed a bug. If you flick to swipe a page and then quickly swipe again in the same direction while it hasn't settled yet, it snaps back to previous position. I'll check this tonight before doing a release.

satya164:
Hmm.. it's not just for the same direction, but for opposite direction too. We'll likely need to add some more checks. I'll take a proper look later since I'm at work right now.

julienvincent:
Hmm. Apologies for missing that - Also at work now and will also check it out this evening.

julienvincent:
I have been able to reproduce what you describe in the currently released version so I do not think it was added by this PR. What I have noticed is that when flicking and the swiping to cancel, the UI jitters before responding. I think there is some race condition occurring (hence the jitter).

julienvincent:
Thinking about it further - I see now how and why the race condition is caused.

As soon as your finger leaves the screen, a process begins in an attempt to transition from `a => b`. If you then try flick from `b => a` or `a => b` again, another process now begins, starting a race. Generally the last swipe made will win the race.

If done properly a double swipe `a => b` will not cause the screen to remain at `a` but rather a UI jitter will happen and then it will continue to `b`. 

If however your second swipe `a => b` did not cover enough distance to cause a transition (maybe you are attempting to reproduce and are swiping too quickly, ie the second swipe does not pass the threshold) then that second swipe's end position is still going to be screen `a`. It was the latter process in the race resulting in it winning, and therefor the screen transitions back to `a`.

Sorry if my description is not very clear. :)

julienvincent:
What I think should happen is if the user touches the screen during a transition, it should terminate. This logic can be described in the `onStartShouldSetPanResponder` callback: If a transition is in progress then `stopAnimation` can be called.

satya164:
> I have been able to reproduce what you describe in the currently released version so I do not think it was added by this PR

Hmm... Nope. It was not there earlier. I tested by removing the new check, and it worked as expected. Probably a GIF will help.

> If however your second swipe a => b did not cover enough distance to cause a transition (maybe you are attempting to reproduce and are swiping too quickly

I'm taking both distance and velocity of swipe into account, so even if my swipe didn't cover enough distance, the velocity of my swipe should be enough to trigger the transition, as I'm swiping quickly

> What I think should happen is if the user touches the screen during a transition, it should terminate

Yeah, I thought `respondToGesture` is enough to handle it, but seems it's not. We can call `position.stopAnimation()` in `startGesture`

I'd look into it a bit more, but I'm too tired today. So will do it tomorrow.

julienvincent:
>Hmm... Nope. It was not there earlier. I tested by removing the new check, and it worked as expected. Probably a GIF will help.

To clarify what I meant - What I thought I could reproduce (To some degree, not entirely) is the UI jitter caused by what I thought were race conditions.

However.. looks like I was dead wrong about the race conditions. I was so sure that's what was happening :/. After a bit of debugging I found that gestureState values (`dx` `dy` `vx` ... ect) are all `0` when swiping during a transition. I have no idea why - might be a bug with react-native or something else entirely. 

Either way this resulted in startDirection being set to 0, and `isReverseDirection` doesn't expect startDirection to ever be `0`.

julienvincent:
Hi again, so I have been further using and testing and the logic added by this pr still is not enough.

There are still a few gesture processes that cause the UI to close or open incorrectly. I will be working on a more complete solution when I get some time. 

Additionally I realize that `onStartShouldSetPanResponder[Capture]` should not be false as it means swiping is not detected when gesture is on top of a `Touchable` component.

satya164:
> I will be working on a more complete solution when I get some time.

Thanks a lot!

> Additionally I realize that onStartShouldSetPanResponder[Capture] should not be false as it means swiping is not detected when gesture is on top of a Touchable component.

I'm not sure about this, this only means the start of the touch is not detected, but when you move your finger `onMoveShouldSetPanResponder` should get called and we should be able to handle the swipe. At least that's how it was in the last project I worked on.

julienvincent:
>I'm not sure about this, this only means the start of the touch is not detected, but when you move your finger onMoveShouldSetPanResponder should get called and we should be able to handle the swipe. At least that's how it was in the last project I worked on.

So I'm still not 100% sure how the intricacies on touch handling works, I am gonna need to spend some time figuring it out completely - but the problem I am experiencing is when trying to swipe over button elements, or Views wrapped in `TouchableWithoutFeedback`. If I set `onStartShouldSetPanResponder[Capture]` to `false` then I can't swipe properly, if its set to `true` then swiping works fine but I can't press the button.

PR #43, opened 2016-10-18
Use Animated.View for TabBar root container

clauderic:
This PR simply replaces the root container of the `TabBar` component from `View` to `Animated.View`. 
This way, properties like `backgroundColor` can be dynamically animated based on the active tab.

I don't think there are any side effects to doing this, but let me know if I'm missing something. 

Cheers!


satya164:
LGTM

PR #45, opened 2016-10-19
Check if startDirection is 0

julienvincent:
This should fix the bug described in #42.


satya164:
Wow. You're on a roll. Thank you.

satya164:
I've published a new release

PR #49, opened 2016-10-20
Added optional swipeEnabled prop to TabViewAnimated.

alvaromb:
You can disable the automatic swipe for a `TabViewAnimated` component.

### DIFF
--- a/README.md
+++ b/README.md
@@ -102,6 +102,7 @@ Container component responsible for managing tab transitions
 It accepts the following props,
 - `navigationState` - the current navigation state
 - `configureTransition` - optional callback which returns a configuration for the transition, return `null` to disable animation
+- `canJumpToTab` - optional callback which accepts a route, and returns a boolean indicating whether jumping to the tab is allowed
 - `onRequestChangeTab` - callback for when the current tab changes, should do the `setState`
 - `onChangePosition` - callback called with position value as it changes (e.g. - on swipe or tab change), avoid doing anything expensive here
 - `shouldOptimizeUpdates` - whether to implement a `shouldComponentUpdate` strategy to minimize updates, enabled by default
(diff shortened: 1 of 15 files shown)


knowbody:
I don't think there is a point for having it as you can already accomplish it via `renderPager`, see NoAnimation exmaple

alvaromb:
Mmm, you're right @knowbody. Just thought it was handy.

alvaromb:
Thanks for your feedback Mateusz! 🙇

knowbody:
no, thanks for your PR! I think you can leave it open and see what @satya164 has to say, I was actually using that yesterday, the first thing I did was to add `swipeEnabled` on `TabViewAnimated`, but then @satya164 told me about `renderPager`. So it would just be a duplication, I think

alvaromb:
I think is better to discard this PR, you're right it makes no sense because you can set the pager through `renderPager`. However, maybe it's interesting to stick an small example into the docs. I will be more than happy to provide a snippet if you guys find it's useful.

BTW, great component!

knowbody:
Definitely useful to document it! mind sending a PR?

satya164:
Thanks a lot for the PR. I think `swipeEnabled` on `TabViewAnimated` is a better API, but since we support swappable pagers, it's not really the correct thing to do. Since the underlying pagers could accept different props (for example, `TabViewPagerPan` accepts 2 extra props), and it might not have a `swipeEnabled` prop at all depending on the implementation. I guess it's just maintaining a balance between customisability and simplicity.

About docs, I think we can have a FAQ section outlining common things like this. Help is appreciated :)

alvaromb:
Ok! I'm in the middle of something, I'll send a PR with a FAQ proposal later today.

PR #58, opened 2016-11-03
Fixes tab transition to jump if update occurs during transition

### LINKED ISSUE #55: a setState during a tab transition will cause a jump
with current way shown in the example, it looks like I have to **control** the tab ~ like this:

```
  onRequestChangeTab={this._handleChangeTab}
...
_handleChangeTab = (index) => this.setState({ index })
```

the problem I see is that _handleChangeTab is always called AFTER the transition is done, but not "before" , and the problem is if one setState happen **or any tree update**, the render() get called again with the **previous tab index** which causes a jump.

I have tried `0.0.37` as well as `master`. Both have the problem.

gre:
Fixes #55

### DIFF
--- a/src/TabViewTransitioner.js
+++ b/src/TabViewTransitioner.js
@@ -75,8 +75,10 @@ export default class TabViewTransitioner extends Component<DefaultProps, Props, 
     }
   }
 
-  componentDidUpdate() {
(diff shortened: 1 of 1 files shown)


satya164:
Thank you!

satya164:
I'll do a release tomorrow.

PR #65, opened 2016-11-14
refactor: inline pan responder functions to TabViewPagerPan

julienvincent:
The direction tracking variable `startDirection` is initialized inside a function that is called per-render. This resulted in direction tracking not being remembered across renders and prevented correct swipe cancel behaviour.

This PR moves `startDirection` out of the`forHorizontal` function and allows its value to carry across renders.


satya164:
I don't like this very much since the variable is no longer tied to the component. I think we can do some refactor and just inline it to `TabViewPanPager` component instead of a separate function and then we can store this variable's in component's instance.

It is a different function for historical reasons which no longer apply to the current situation.

julienvincent:
Alright we can do that instead, but it is important that startDirection is initialized once _only_.

satya164:
@julienvincent You mean once per component, right?

julienvincent:
Yes, once per component.

src/TabViewPanResponder.js
```diff
@@ -41,8 +41,6 @@ function forHorizontal(props: Props) {
       return gestureState.vx < 0;
     } else if (startDirection < 0) {
       return gestureState.vx > 0;
-    } else {
```
satya164: @julienvincent But it's still returning a falsy value, so nothing will actually change here, except you get eslint error.

julienvincent: @satya164 True nothing will change, but it was an unnecessary check. I will change the check to something like this instead:

```
if (vx < 0) {

} else {

}
```

julienvincent:
I will build a basic implementation of having `startDirection` in `TabViewPanPager` to see if everything works as expected, and then we can restructure properly if everything works

satya164:
@julienvincent great!

julienvincent:
@satya164 Done, seems to work as expected.

src/TabViewPagerPan.js
```diff
@@ -121,8 +122,17 @@ export default class TabViewPagerPan extends Component<DefaultProps, Props, void
...
+    const shouldCapture = this._callResponderMethod('onMoveShouldSetPanResponder', false)(evt, gestureState);
+    if (shouldCapture) {
+      this.startDirection = gestureState.dx;
+    }
+    return shouldCapture;
+  }
+
   _panHandlers: any;
   _panResponder: any;
+  startDirection = 0;
```
satya164: nit: start direction

src/TabViewPanResponder.js
```diff
@@ -36,13 +35,11 @@ function forHorizontal(props: Props) {
     );
   }
 
-  function isReverseDirection(gestureState: GestureState) {
+  function isReverseDirection(gestureState: GestureState, startDirection: number) {
```
satya164: I think it's better if we move all the functions there and delete this file

julienvincent:
@satya164 Done. I also underscorified all function names that were imported. Assuming that is what you want?

satya164:
@julienvincent Yeah. Thanks a lot!

julienvincent:
@satya164 np 👍

src/TabViewPagerPan.js
```diff
@@ -52,77 +55,112 @@ export default class TabViewPagerPan extends Component<DefaultProps, Props, void
...
+      const nextIndex = index - (gestureState.dx / Math.abs(gestureState.dx));
+      if (this._isIndexInRange(nextIndex)) {
+        return nextIndex;
+      }
+    }
+    return index;
+  };
+
+  _canMoveScreen = (evt: GestureEvent, gestureState: GestureState) => {
+    const { navigationState: { routes, index } } = this.props;
```
satya164: I forgot this, here you should add,

```js
if (this.props.swipeEnabled === false) {
  return false;
}
```

Otherwise the prop is not used anywhere.

julienvincent: Ah! My bad completely missed that. Nice catch.

PR #69, opened 2016-11-15
fix maximum callstack size error in jumpToTab

knowbody:
this fixes jumpToTab being called on the unmounted component


satya164:
@knowbody you didn't add the whole diff! it should set the value of `this._mounted` in `componentDidMount` and `componentWillUnmount`

knowbody:
I'll fix it tomorrow. Don't merge it until then

PR #119, opened 2017-02-10
Add "accessible" and "accessibilityLabel" props into <TouchableItem />

isnifer:
Hi, guys!

This PR brings ability to use accessibilityLabel on Android while E2E-testing


satya164:
Thanks a lot!

Issue #130, opened 2017-02-15
Tutorial on adding tabs dynamically


spezzino:
Is there a way to load tabs dynamically?
For example, I call a service and then populate the routes array with every tab I want to show.

satya164:
@spezzino I've never tried that, but it should be possible by just updating the `routes` array in the state. if it doesn't work properly happy to investigate why and fix it. though what I had done in a previous app was just show a loading spinner without the tabview and then when I get the list, show the tabview with the new set of tabs.

spezzino:
It worked.

In case anyone needs an example, I'll do a PR in a moment to your `example` project

satya164:
@spezzino sounds good

CharlesMangwa:
Problem seems to have been solved here, closing 👍

brendandestefano:
@spezzino did this tutorial ever happen? looking for dynamic tabs at the moment

lcardosozago:
an example would be nice

pratikkarad:
@satya164 have you uploaded loading dynamic tab example?
@brendandestefano did you find any solution on dynamic tabs thing?
@spezzino after executing it's giving me an error, if you can provide any working example on expo, that would be very great

PR #131, opened 2017-02-16
Example for dynamic tab load

spezzino:


karanjthakkar:
@spezzino I was looking into #130 to understand the motivation behind this PR. I think the way your example portrays dynamic tab loading is essentially just showing a loading indicator and then rendering the actual TabView. Personally, this doesn't seem to be a solution specific to this library but just how you load any component that depends on some data that may be available in the future. That being said, I'll leave it to @satya164 to see if he believes this might add value. Thanks for your effort!

spezzino:
It is just an example to help people looking into this solution

satya164:
I'll add a better example soon.

dsp1589:
@satya164 did you create better example? It re-renders all the tabs. one of the tab contains webview which is reloading every time we add a tab.

khat33b:
I tried running this example. I am getting this error:
```
Invariant Violation: Invariant Violation: Element type is invalid: expected a string (for built-in components) or a class/function (for composite components) but got: undefined. You likely forgot to export your component from the file it's defined in, or you might have mixed up default and named imports.
```

PR #141, opened 2017-02-22
Maintain backward compatibility  for react native 0.39

mattsp:
Into the TabViewPagerScroll component we have a property keyboardShouldPersistTaps wich take a boolean property in react native 0.39 and not a string look like react native 0.4x.

I have add a new property for configure this property if we need it to change and fallback value


mattsp:
What is the prolbeme with my pull request

satya164:
I don't think adding a prop for backward compatibility is a good way. For now you can fork the lib, or we could publish a modified version to npm for backward compatibility.

PR #145, opened 2017-02-23
Guard scrollTo call so that Jest completes

c-h-:
When testing with Jest in my project I get `TypeError: _this._scrollView.scrollTo is not a function`. Add this simple condition makes all my tests pass. If there's a better way to fix this that would be appreciated!


satya164:
Can you file a bug report in React Native's repo? This is a bug in jest's React Native preset. I don't want to add this check because it just suppresses the issue. The tests should behave the same as actual code.

PR #146, opened 2017-02-24
Add support for activeBackgroundColor for focused tab on Android

coodoo:
See it in action [here](https://www.dropbox.com/s/1n06oqpz985xhky/a3.mov?dl=0)

Main use case is when on iOS and putting the tabBar on top, I need it to scroll horizontally and also have focused tab highlighted (with `activeBackgroundColor` style), hence the PR.


satya164:
You can set `tabStyle` according to the state.

PR #163, opened 2017-03-09
Fix checking of scrollTo()

zvona:
This error occurs with Jest and tests won't pass.

### DIFF
--- a/src/TabViewPagerScroll.js
+++ b/src/TabViewPagerScroll.js
@@ -64,7 +64,7 @@ export default class TabViewPagerScroll extends PureComponent<void, Props, void>
   _isMomentumScroll: boolean = false;
 
   _scrollTo = (x: number) => {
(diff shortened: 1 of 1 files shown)


satya164:
Can you file a bug report in React Native's repo? This is a bug in jest's React Native preset. I don't want to add this check because it just suppresses the issue. The tests should behave the same as actual code.

zvona:
Ok, thanks for reply. I examined a little and this really is a generic issue on Jest. Both ListView and ScrollView needs to be mocked in order to run the tests properly.

https://github.com/facebook/react-native/issues/11659#issuecomment-282448969

alvaromb:
I've just opened a bug in the `react-native` repo https://github.com/facebook/react-native/issues/13034

alvaromb:
PR: https://github.com/facebook/react-native/pull/13048

Issue #159, opened 2017-03-06
will web platform be added?


agrcrobles:
Hey,

I think that there is no much to add in order for the web platform to be added, but styles and strip flow off on prepublish.

will you consider adding it or accept PRs?

satya164:
Surprised it works on the web without modification. Have you tried it? Do you have a demo?

agrcrobles:
yes! that is awesome.
The pager that works fine is `TabViewPagerPan`.

![ehlef8](https://cloud.githubusercontent.com/assets/5727373/23618884/2659baa2-0292-11e7-9c1e-e14ff9f5e10f.gif)
![j5y6yj](https://cloud.githubusercontent.com/assets/5727373/23618889/2a3d03fe-0292-11e7-8566-d9addbade16a.gif)

Probably it is better to have a repo to test it but in the meantime here goes some comments.

Scroll style in [TabBar](https://github.com/react-native-community/react-native-tab-view/blob/9eae4cc00ffec888a03cae6c18131f908e1f3595/src/TabBar.js#L20) works better `auto` than `scroll`. ( prevent scrolling below. )

![screenshot from 2017-03-06 17-36-37](https://cloud.githubusercontent.com/assets/5727373/23619330/a05fefe6-0293-11e7-9cda-caaecba7d385.png)

Also found that width offset should be recalculated when browser resizes ( not sure how it works when device orientation change ).

well that is all, hope this helps @satya164

satya164:
That's exciting! Let's add a prepublish script for web as the first step. Regarding `overflow`, it didn't work without `scroll` in native apps, so we'll need to check that.

agrcrobles:
yes it is!
I have just created https://github.com/agrcrobles/react-native-tab-view-web and moved some examples there.
There is need to include node_modules folder in [webpack.config](https://github.com/agrcrobles/react-native-tab-view-web/blob/master/webpack.config.js#L26)

brunolemos:
@satya164 it would be better if this packaged generated a `dist/` folder ES5 compilent, so we dont have to `eject` from `create-react-app` to edit webpack config or use a fork.

satya164:
Yeah, please send a PR adding the script and I'll review.

CharlesMangwa:
Closing for now, could be re-opened when another PR will be submitted.

CharlesMangwa:
Closed #159.

agrcrobles:
[...]

PR #166, opened 2017-03-10
Removed scroll on header in web

agrcrobles:
The same discussed in #159

### DIFF
--- a/src/TabBar.js
+++ b/src/TabBar.js
@@ -18,7 +18,7 @@ const styles = StyleSheet.create({
     flex: 1,
   },
   scroll: {
-    overflow: Platform.OS === 'web' ? 'none' : 'scroll',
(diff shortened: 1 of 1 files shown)


src/TabBar.js
```diff
@@ -17,7 +18,7 @@ const styles = StyleSheet.create({
     flex: 1,
   },
   scroll: {
-    overflow: 'scroll',
+    overflow: Platform.OS === 'web' ? 'none' : 'scroll',
```
satya164: `none` is not a valid value for overflow on web, maybe you meant `auto`?

Valid values are `visible`, `hidden`, `scroll` and `auto`.

agrcrobles: My bad, it was auto, I will update it.

PR #167, opened 2017-03-10
Transpile lib for web

agrcrobles:
Discussed in #159 

Overall:

 - preserved flow intellisense in native apps.
 - worked on web. ( not tried flow intellisense )
 - lost source map.
 - yarn needs to be run

Needless to say need to be tested.


agrcrobles:
not sure if a polyfill fits in better, I will close this in the meantime.

PR #175, opened 2017-03-22
Add `containerStyle` prop to TabViewAnimated

mikehobi:
First off, thank you for this project, it's great!

I've run into this issue with having to overwrite the default `styles.container` style that is used in `TabViewAnimated`'s `_renderItems()` function. I am using it as a carousel that's inline with other content, and does not need to be full height. I can pass a height to the component, but in my case, I'd rather have the content dictate the height and pass `{ flexGrow: 0 }`.

Would love some feedback!

### DIFF
--- a/src/TabViewAnimated.js
+++ b/src/TabViewAnimated.js
@@ -28,6 +28,7 @@ type Props = TransitionerProps & {
   renderHeader?: (props: SceneRendererProps) => ?React.Element<*>;
   renderFooter?: (props: SceneRendererProps) => ?React.Element<*>;
   lazy?: boolean;
+  containerStyle?: any;
 }
 
 type State = {
@@ -57,6 +58,7 @@ export default class TabViewAnimated extends PureComponent<DefaultProps, Props, 
     renderFooter: PropTypes.func,
     onChangePosition: PropTypes.func,
     lazy: PropTypes.bool,
+    containerStyle: View.propTypes.style,
   };
 
   static defaultProps = {
@@ -86,11 +88,11 @@ export default class TabViewAnimated extends PureComponent<DefaultProps, Props, 
   };
 
   _renderItems = (props: SceneRendererProps) => {
-    const { renderPager, renderHeader, renderFooter } = this.props;
+    const { renderPager, renderHeader, renderFooter, containerStyle } = this.props;
     const { navigationState } = props;
 
     return (
-      <View style={styles.container}>
+      <View style={[ styles.container, containerStyle ]}>
         {renderHeader && renderHeader(props)}
         {renderPager({
           ...props,


src/TabViewAnimated.js
```diff
@@ -86,11 +88,11 @@ export default class TabViewAnimated extends PureComponent<DefaultProps, Props,
   };
 
   _renderItems = (props: SceneRendererProps) => {
-    const { renderPager, renderHeader, renderFooter } = this.props;
+    const { renderPager, renderHeader, renderFooter, containerStyle } = this.props;
     const { navigationState } = props;
 
     return (
-      <View style={styles.container}>
+      <View style={[ styles.container, containerStyle ]}>
```
satya164: Let's name it as just `style`

satya164:
I think it'll be better to get rid of the wrapping view. We can just return an array and the styles we have for the wrapping view in `TabViewTransitioner`

mikehobi:
That sounds like a much cleaner solution. Nice to get rid of unnecessary Views.

My guess is that it would look something like this?
```jsx
return [
  renderHeader && renderHeader(...),
  renderPager(...),
  renderFooter && renderFooter(...),
]
```

What would be the best way to handle keys? 

If you passed it: `renderHeader && renderHeader({...props, key: 'header'})`, that would require the `renderHeader` function to either always spread props or explicitly pass the key`<TabHeader key={props.key} />`, correct?

satya164:
Yeah.

I think `renderHeader({...props, key: 'header'})` looks fine.

mikehobi:
Thinking about this a little more. The `renderHeader({...props, key: 'header'})` approach would give users pretty ambiguous errors if they did not pass the key down:

```
Each child in an array or iterator should have a unique "key" prop.
Check the render method of `TabViewTransitioner`
```

 ☝️ This would be the error if a user's `renderHeader` function looked like this:
```jsx
renderHeader(props) {
  const { index, routes } = props.navigationState
  return (
    <TabHeader 
      routes={routes}
      currentIndex={index}
    />
  )
}
```
Kind of a weird side affect of this abstraction and might not be the most obvious warning for users unless it was documented

satya164:
There are several ways to solve this issue,

1) We can manually add the key using `React.cloneElement`
2) Instead of using an array in `TabViewTransitioner`, refer them like `{items[0]}` etc.
3) Remove the wrapping view in `TabViewTransitioner`, something like,

```js
return this.props.render({ ...this._buildSceneRendererProps(), onLayout: this._handleLayout });
```

I'm not too worried about it right now because `TabViewTransitioner` will be removed pretty soon when https://github.com/react-native-community/react-native-tab-view/tree/native-animation is ready

mikehobi:
Alright thanks! I'm stoked for the native animation.

I'll close this Pull Request, and see if I can help make any progress on what you have suggested.

PR #190, opened 2017-04-01
TabHeader styling

rafi-isakh:
In the recent source code, the style for tabHeader style is hard-coded. I need to change the tabHeader background color. So, I add a property to make tabHeader style customizeable


satya164:
You can pass `style` prop to change background color.

Issue #184, opened 2017-03-27
RTL is not supported!


moerabaya:
What i'm suffering from is that the tabs still functioning the same as the LTR after switching to RTL, I know what seems to be the issue because I have a custom indicator, but i'm not able to use negative values to translate them to the opposite.

Thank you so much.

![untitled](https://cloud.githubusercontent.com/assets/12199727/24356535/b80fb316-12fa-11e7-9a82-14c41bff94e9.gif)

satya164:
Can you send a PR to fix this?

moerabaya:
@satya164 I never send one, i'm not sure how to do that?

yaronlevi:
This bug also affects react-navigation. The TabNavigator is not working:

https://github.com/react-community/react-navigation/issues/860

yaronlevi:
Inside TabViewPagerScroll.js putting `flexDirection: 'row-reverse'` on the `<ScrollView />` fixes the problem partially. The tab's contents is shown but the order is reverse.

dastoori:
#https://github.com/facebook/react-native/issues/11960

dastoori:
@yaronlevi, look at my last comment here: https://github.com/facebook/react-native/issues/11960#issuecomment-292732636

@moerabaya, I sent a PR #194, until it merged, you can use one of these:
- `npm i -S https://github.com/dastoori/react-native-tab-view.git#rtl-support`
- `npm i -S https://github.com/dastoori/react-navigation.git#rtl-support`

satya164:
Re-opening this since RTL is not yet supported in `TabViewPagerScroll`, hence unsupported on iOS.

dastoori:
@satya164, iOS ScrollView has been fixed in ReactNative v0.43.2 (https://github.com/facebook/react-native/issues/11960#issuecomment-292732636)

satya164:
Nice. I think we may still need to handle the scroll position etc.

dastoori:
I checked this in version `0.0.59`, i used the examples in `examples` folder. These examples does not works properly:

- "1. Scrollbar top bar"

![scrollabletopbar](https://cloud.githubusercontent.com/assets/2003143/25232217/ca6c2e10-25ef-11e7-82ea-ba38419d9549.gif)

- "3. Bottom bar with indicator"

[...]

PR #194, opened 2017-04-08
RTL support

dastoori:
Fix for #184


satya164:
Thanks a lot for the PR! Sucks that `ViewPagerAndroid` doesn't handle RTL by default.

PR #204, opened 2017-04-18
Fix state change issue in RTL layout

dastoori:
I found that in RTL layout when i swipe between some tabs, the `navigationState.index` doesn't updates (it doesn't calls `props.jumpToIndex()`)

This is related to my last PR (#194),


satya164:
Thank you for the PR. Can you try master and see if the issue still exists?

dastoori:
@satya164, I checked, and it works perfectly fine on both Android and iOS.

Thanks for the release, Can you update the version in the `react-navigation`?

satya164:
@dastoori would be great if you can send a PR :D

PR #207, opened 2017-04-21
keyboardShouldPersistTaps rn 0.38 bool, and swipeEnabled prop support

monkwooo:
rn 0.36 requires keyboardShouldPersistTaps to be boolean,
also
swipeEnabled prop was not passed to native tabview from TabViewTransitioner


satya164:
Latest RN expects a string.

Regarding `swipeEnabled`, `TabViewTransitioner` doesn't accept this prop. You have to directly pass it to the pager. The API will be revised in a future version.

PR #210, opened 2017-04-22
chore: move eslint-plugin-prettier to devDependencies

fson:
I noticed this dependency was added to the dependencies of my app after upgrading to the latest version. However, it's only needed during development, so it can be moved to `devDependencies`.


satya164:
Thank you!

PR #211, opened 2017-04-22
Update GitHub URLs in package.json

fson:


fson:
Uh, just noticed also the name of the repo had changed from `react-native-tab-view` to `react-native-tabview`. Sorry for missing that in this PR – fixed in #212.

PR #220, opened 2017-04-27
fix: try to update scroll only when idle

satya164:


jasan-s:
At Initial Load , Tap is unresponsive in changing to different tabViews.  Requires Swiping to another tab and then the tapping is functional.

PR #230, opened 2017-05-03
Fix Flow errors that appear in React Native master and Flow 0.45.0

Ashoat:
I have reason to use React Native master (and corresponding Flow 0.45.0) on a personal project, and I'm seeing some Flow [errors](https://gist.github.com/Ashoat/95569ccc5dd40b1c5ff9d0f5d59b461a) when used with the latest react-navigation package. 6/8 of the Flow errors are fixed in this [pull request](https://github.com/react-community/react-navigation/pull/1365) on react-navigation. The last two errors are addressed here.

One fixed issue is with the custom defined Style prop. Before [this fix](https://github.com/facebook/react-native/commit/c7387fefc85fd15dceadfa291857099ed3e848b8) landed last month, Flow wasn't evaluating the types of any imports from react-native. This fix isn't in RN 0.44.0, but it is in master.

As a result, the mismatch because the custom specification of the Style prop here and the specification in the `react-native` package is now causing a Flow error. I fixed the issue by simply aliasing the `StyleObj` from `react-native`. Somewhat more details available [here](https://github.com/flowtype/flow-typed/issues/631#issuecomment-298982849).

The second fixed issue is that `View.propTypes` is now deprecated in favor of `ViewPropTypes`. That change is [here](https://github.com/facebook/react-native/commit/53905a537a58d028fc5eb1c4e9a8ec967d8cd396).


src/TabViewTypeDefinitions.js
```diff
@@ -58,4 +60,4 @@ export type PagerProps = {
   children?: *,
 };
 
-export type Style = Object | number | false | null | void | Array<Style>;
+export type Style = StyleObj;
```
satya164: So this type `number | bool | Object | Array<?Atom>` seems incorrect because style doesn't actually accept a `boolean` (i.e. accepts `false`, but not `true`), but I guess we can't do anything here unless that's fixed. :(

src/TabBar.js
```diff
@@ -70,7 +71,7 @@ export default class TabBar<T: Route<*>>
     renderLabel: PropTypes.func,
     renderIndicator: PropTypes.func,
     onTabPress: PropTypes.func,
-    tabStyle: View.propTypes.style,
+    tabStyle: ViewPropTypes.style,
```
satya164: This is an unfortunate breaking change. I think we might need to remove this prop type entirely for now :(

Ashoat: What's it breaking?

satya164: This is only present in RN 0.44, this breaks the library for all older versions.

Ashoat: You are right. I'll go ahead and remove these `PropTypes` then, but here's an alternate approach for your consideration that I tested on RN 0.43.4:

```js
import {
  View,
  ViewPropTypes,
} from 'react-native';
...
    tabStyle: ViewPropTypes ? ViewPropTypes.style : View.propTypes.style,
```

The destructuring import assigns undefined if the requested module does not exist in the specified file. Meanwhile, Flow is crippled on 0.43.4 by [this aforementioned problem](https://github.com/facebook/react-native/commit/c7387fefc85fd15dceadfa291857099ed3e848b8), so it doesn't complain about the misnamed import.

PR #233, opened 2017-05-04
Fix tabBarVisible on android device.

mindk:
Fix bug with blank screen on android device when set  tabBarVisible = true.

### DIFF
--- a/src/TabViewAnimated.js
+++ b/src/TabViewAnimated.js
@@ -263,7 +263,9 @@ export default class TabViewAnimated<T: Route<*>>
         loaded={this.state.loaded}
         style={[styles.container, this.props.style]}
       >
-        {renderHeader && renderHeader(props)}
+        <View style={styles.topBar}>
+          {renderHeader && renderHeader(props)}
+        </View>
         {renderPager({
           ...props,
           ...rest,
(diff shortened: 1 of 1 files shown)


jsierles:
Why is this opacity value required for Android?

satya164:
Why does this work? Do you have link to an issue in RN repo?

mindk:
Android can't rerender screen with animation if we remove one of elements before content (like a top bar container). But it works correctly if we change dimensions of this elements. So I wrapped topBar into View element. View with style {opacity: 1} works as separately container, but  if remove this style than we will have same bug again.

satya164:
Not sure what you mean by that. The disappearing screen bug seems to be caused by buggy zIndex implementation on Android (https://github.com/react-native-community/react-native-tab-view#tabbar-disappears-or-shows-weird-behaviour-when-conditionally-rendering-items).

Your PR seems to fix that, because `opacity: 1` implicitly sets the wrapper view to be non-collapsable on Android, and the wrapper `View` means that the tabbar is no longer the sibling of the view, which results in the z-Index related bug not appearing.

I just pushed a commit which explictly sets `collapsable` to `false`, which should fix the issue -
 c80fdff7bdb59b281c324bb6fb0dd1f213c1aae6

PR #236, opened 2017-05-07
Remove all references to View.propTypes

Ashoat:
Follow-up to #230: for compatibility with both RN 0.44.0 as well as earlier versions, we decided there to just omit all of these PropTypes.

Besides just removing these PropTypes, here is an alternate approach we can take here to preserve compatibility across RN versions:

```javascript
import {
  View,
  ViewPropTypes,
} from 'react-native';
...
tabStyle: ViewPropTypes ? ViewPropTypes.style : View.propTypes.style,
```
The destructuring import assigns undefined if the requested module does not exist in the specified file. Meanwhile, Flow is crippled on 0.43.4 by the problem [this commit fixed](https://github.com/facebook/react-native/commit/c7387fefc85fd15dceadfa291857099ed3e848b8), so it doesn't complain about the misnamed import. I've tested this on RN 0.43.4, RN 0.44.0, and master.

### DIFF
--- a/src/TouchableItem.js
+++ b/src/TouchableItem.js
@@ -46,9 +46,8 @@ export default class TouchableItem
   };
 
   render() {
-    /* eslint-disable react/prop-types */
+    // eslint-disable-next-line react/prop-types
     const { style, pressOpacity, pressColor, borderless, ...rest } = this.props;
-    /* eslint-enable react/prop-types */
 
     if (Platform.OS === 'android' && Platform.Version >= LOLLIPOP) {
       return (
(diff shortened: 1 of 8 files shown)


satya164:
Thank you for the PR. CI is failing due to the missing proptype. We should suppress the error.

src/TouchableItem.js
```diff
@@ -46,7 +46,9 @@ export default class TouchableItem
   };
 
   render() {
+    /* eslint-disable react/prop-types */
```
satya164: `// eslint-disable-next-line react/prop-types` is nicer

PR #238, opened 2017-05-09
fix loop call problem when using with react-native-web

jefferscn:
I'm working on a project using react-native-web & react-native-tab-view,everything works excellent,util when I change tab fast using click.There is a loop call , when Animiated.start callback executes, navigationState.index has changed, then another onRequestChangeTab will be called;

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "react-native-tab-view",
-  "version": "0.0.64",
+  "version": "0.0.61",
   "description": "Tab view component for React Native",
   "main": "src/index.js",
   "files": [
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "react-native-tab-view",
-  "version": "0.0.61",
+  "version": "0.0.64",
   "description": "Tab view component for React Native",
   "main": "src/index.js",
   "files": [
--- a/src/TabViewAnimated.js
+++ b/src/TabViewAnimated.js
@@ -200,7 +200,7 @@ export default class TabViewAnimated<T: Route<*>>
     };
   };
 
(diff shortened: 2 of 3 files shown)


satya164:
Can you explain what the change does? Looks like you never update the state when the animation finished and `jumpToIndex(index, false)` is basically the same as not calling `jumpToIndex` at all.

If there is a race condition, then we need to handle the race condition in the pager.

jefferscn:
There are two ways to change tab.
- onClick
  navigation change lead to an animation,
  when animation finished,jumpToIndex should not be called.

- onSwipe
  when animation finished,jumpToIndex will be called, and lead to navigation change.

On native usage, TabViewPagerPan is not used,so I cannot test it;
But I test it in web application using react-native-web & react-native-tab-view,in this
situation,when click too fast,there would be two(or more) animations running at the same time.
When one animation finished,jumpToIndex will be called,in normal way,`index !== navigationState.index` should return true,but Unfortunately another click has change the navigationState.index,then onRequestChangeTab is called,anothing animation will happen.

satya164:
> On native usage, TabViewPagerPan is not used,so I cannot test it

You can test it by using `renderPager` prop 

So from what I understand, according to your changes, you only update the state when the gesture finishes and not when the animation occurs due to state change. I would prefer if you don't call `jumpToTab` at all instead of passing a n extra argument which decides if state should be updated.

However, the actual issue is that there is a race condition here, and when one animation is cancelled by a new animation, the old one still tries to update the state. The fix will be to update the state only in the latest animation and not in cancelled animations. If you try the latest release, it should already work this way. Can you try that and let me know?

jefferscn:
Well done.Working!

PR #254, opened 2017-05-17
Add high zIndex to focused child in TabView's

skevy:
In react-navigation, we have a shadow on the edge of a card inside a CardStack. This shadow is usually offscreen. However, when you have two tabs, and a CardStack in each tab, the following happens (the shadow has been turned blue and increased in size to better demonstrate the issue):

sample

This PR adds a high zIndex to the focused child in TabViewPagerPan and TabViewPagerScroll, thus making sure the focused child is on top.

### DIFF
--- a/src/TabViewPagerPan.js
+++ b/src/TabViewPagerPan.js
@@ -281,19 +281,21 @@ export default class TabViewPagerPan<T: Route<*>>
         ]}
         {...this._panResponder.panHandlers}
       >
-        {Children.map(children, (child, i) => (
-          <View
-            key={navigationState.routes[i].key}
-            testID={navigationState.routes[i].testID}
-            style={
-              width
-                ? { width }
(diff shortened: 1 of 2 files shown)


satya164:
Hey! Are you facing this with the master branch of react-navigation? This should take care of the shadow:
https://github.com/react-community/react-navigation/blob/master/src/views/TabView/TabView.js#L69
https://github.com/react-community/react-navigation/blob/master/src/views/TabView/TabView.js#L198

I want to avoid adding z-Index here since z-Index is buggy on Android and will cause tabs to be messed up when a new tab is added/removed dynamically. Doesn't matter for `react-navigation` though since the number of tabs is constant. Though not sure why `overflow: hidden` doesn't work? definitely worked when I tested it.

satya164:
Closing this since we talked about it.

PR #270, opened 2017-05-30
Revert "fix: remove unnecessary flexGrow"

michaelduminy:
This reverts commit 48741e7ea77d22e0b010a94d7985d186cf565539.

See discussion for details https://github.com/react-community/react-navigation/pull/1713#issuecomment-304831514

### DIFF
--- a/src/TabViewPagerPan.js
+++ b/src/TabViewPagerPan.js
@@ -301,7 +301,7 @@ export default class TabViewPagerPan<T: Route<*>>
 
 const styles = StyleSheet.create({
   sheet: {
-    flex: 1,
+    flexGrow: 1,
     flexDirection: 'row',
     alignItems: 'stretch',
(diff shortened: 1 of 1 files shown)


satya164:
This was added to fix https://github.com/react-community/react-navigation/issues/1248

Reverting it is not a solution.

michaelduminy:
@satya164 Apologies, I wasn't aware of that issue.

I'm not sure where this leaves my issue though. Perhaps I need to understand the internals of `flexGrow` and how that actually fixed the `zIndex` java error on an absolutely positioned top TabBar.

satya164:
@michaelduminy this seems to be a RN issue, if you have an isolated repro, then we can file it in RN repo.

Also you can try latest RN RC, where @janicduplesses redid the zIndex implementation.

PR #280, opened 2017-06-15
fix: prevent android tab view pager from entering inconsistent state

ricmatsui:
- Due to the `requestAnimationFrame` logic, we are seeing `TabBarPagerAndroid` be stuck on the incorrect tab intermittently (as high as 50%)
- We see that `navigationState.index` is being passed correctly to `TabBarPagerAndroid`, but `TabBarPagerAndroid` is showing the incorrect tab
- This is because of a race condition between the existing `requestAnimationFrame` callback and the call to `_handlePageChange ` in `componentDidUpdate`.

- When the issue happens, this is the series of events:

1. `componentWillReceiveProps` is called with `nextProps` with `navigationState.index = 0`
2. `requestAnimationFrame` is called with callback to set page to `0` because of new child count
3. `componentDidUpdate` is called with `navigationState.index = 0` and page is set to `0`
4. Navigation state changes with `react-navigation`
5. `componentWillReceiveProps` is called with `nextProps` with `navigationState.index = 1`
6. `requestAnimationFrame` is *not* called
7. `componentDidUpdate` is called with `navigationState.index = 1` and page is set to `1`
8. Callback for `requestAnimationFrame` from (2) is called and page is set back to `0`

This means that the `TabViewPager` is on page `0` even though `navigationState.index = 1`.

- To prevent this, we do not call the callback for `requestAnimationFrame` if the page has already been set after an animation frame request and the layout and children have not changed
- We also update the callback and don't call `requestAnimationFrame` again if there is a animation frame request already pending

### DIFF
--- a/src/TabViewPagerAndroid.js
+++ b/src/TabViewPagerAndroid.js
@@ -44,7 +44,7 @@ export default class TabViewPagerAndroid<T: Route<*>>
       this.props.layout !== nextProps.layout ||
       Children.count(this.props.children) !== Children.count(nextProps.children)
     ) {
-      global.requestAnimationFrame(() => {
+      this._animationFrameCallback = () => {
         if (this._viewPager) {
           const { navigationState } = nextProps;
           const page = I18nManager.isRTL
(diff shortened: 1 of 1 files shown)


satya164:
Thank you!

satya164:
Haven't got time to test this, but will test before releasing.

jsierles:
I noticed while testing this that swiping stops working. Moving a tab by dragging works, but the swipe gesture no longer works.

PR #301, opened 2017-07-17
Allow position prop for more control over behavior

kkemple:
Hey there! 

We are testing out swipeable tab implementations for our upcoming mobile app at @majorleaguesoccer and we really love this library! It works great right out of the box but we have a few situations in which we don't want the tab bar to be directly connected to the tab views due to intermediate content. We still want to sync tab behavior with tab view behavior because it allows for really nice animations like in the following example:

jul-17-2017 14-24-05
___

What that doesn't show is that the tab bar is in a collapsible header so it can't actually be rendered by the `TabViewAnimated` component:

jul-17-2017 14-04-05
___

My proposed solution is to allow users to pass in their own `postition` prop and providing a default if they don't. This inverts the control and allows the implementor to have full programmatic control over the positioning of tabs and also opens the door for syncing other animations with the views etc.

Im curious to hear your thoughts on this proposal!

### DIFF
--- a/src/TabViewAnimated.js
+++ b/src/TabViewAnimated.js
@@ -16,6 +16,11 @@ import type {
 } from './TabViewTypeDefinitions';
 
 type DefaultProps<T> = {
+  position: Animated.Value,
+  initialLayout: {
+    width: number,
+    height: number,
+  },
   renderPager: (
     props: SceneRendererProps<T> & PagerProps,
   ) => React.Element<any>,
@@ -35,6 +40,7 @@ type Props<T> = PagerProps & {
   renderFooter?: (props: SceneRendererProps<T>) => ?React.Element<any>,
   lazy?: boolean,
   style?: Style,
+  position: Animated.Value,
 };
 
 type State = {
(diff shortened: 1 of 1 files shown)


satya164:
Hello. Thank you for the PR.

I've been experimenting with some ways to implement native animations which will need control over how the `position` value is created and the prop will be unsupported. It's not guaranteed, but adding this prop will likely limit me in what I can do.

Without a `position` prop, there are two ways to achieve what you want,

1. Use the `onChangePosition` callback to update a value passed from parent with `setValue` - While this is the simplest, the behaviour of this may change slightly in future. It's fine for now though
2. Use the Portal pattern where you render an empty component in `renderHeader`, which actually updates a component somewhere in the parent

I hope these are good enough alternatives for you.

kkemple:
i didn't even catch onChangePosition, that will definitely suffice for now, just out of curiosity, what is going to change in the future where that solution will no longer work?

satya164:
@kkemple The signature of `onChangePosition` might change or it might be replaced with something else. Too early to say now :)

PR #303, opened 2017-07-18
Add onSwipeStart and onSwipeEnd props to TabViewPagerPan

### LINKED ISSUE #289: Coverflow example (question)
Hey, is it possible to get the coverflow example to work with TabViewPagerScroll instead of TabViewPagerPan? The issue with the latter I had is that it was inside a Scrollview and swiping sometimes triggered (vertical) scrolling. TabViewPagerScroll doesn't have that issue (due to directionalLockEnabled I guess), but it only shows one item at a time, and I'd like to have the same effect as the coverflow example (showing part of the previous/next item).

Vlad-Zhukov:
There is a bug in React Native that makes gestures to conflict when a PanResponder is put inside of a ScrollView. The workaround for this is to listen to PanResponder start and finish events and toggle ScrollView's `scrollEnabled` accordingly.

Fixes #289

### DIFF
--- a/README.md
+++ b/README.md
@@ -115,6 +115,8 @@ Pager component based on `PanResponder`.
 - `swipeEnabled` - whether to enable swipe gestures
 - `swipeDistanceThreshold` - minimum swipe distance to trigger page switch
 - `swipeVelocityThreshold` - minimum swipe velocity to trigger page switch
+- `onSwipeStart` - optional callback when a swipe gesture starts
+- `onSwipeEnd` - optional callback when a swipe gesture ends
 - `children` - React Element(s) to render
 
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -115,6 +115,8 @@ Pager component based on `PanResponder`.
 - `swipeEnabled` - whether to enable swipe gestures
 - `swipeDistanceThreshold` - minimum swipe distance to trigger page switch
 - `swipeVelocityThreshold` - minimum swipe velocity to trigger page switch
+- `onStartGesture` - optional callback when the gesture starts
```
Vlad-Zhukov: Thanks, fixed

PR #307, opened 2017-07-19
Custom Nav Tabs Width

michelarteta:
```
_renderHeader = props => {
    console.log(props)
    return (
      <TabBar
        {...props}
        scrollEnabled
        style={styles.tabbar}
        tabStyle={styles.tab}
        labelStyle={styles.label}
        tabWidths={[180, 140, 40, 220, 200]}
      />
    );
  };
```
screen shot 2017-07-19 at 5 17 56 pm
changes on stylle tab nav


satya164:
I cannot merge this since it'll not play nice with the indicator. You can implement a custom tab bar for this use case which is pretty straightforward.

PR #318, opened 2017-08-02
Make depreciation warning functional

ixje:


ixje:
The depreciation warning for `onRequestChangeTab` was located after calling `this.props.onIndexChange(index)`. As a result you'll never see the warning because it will error on calling `this.props.onIndexChange(index)` first.

satya164:
Damn. Thank you.

PR #321, opened 2017-08-03
Add prop to disable autofocus on page change or tab tap

fantattitude:
I suggest a property to disable the current default of automatically focusing on page change (either tap in tabBar or swipe when swipe is enabled).


satya164:
Do you really need this behaviour in an app?

satya164:
Closing due to no reply

PR #322, opened 2017-08-03
Add property to TabBar to custom focused tab's label

fantattitude:
Another suggestion which makes it easier to customize focus appearance of tab bar item labels, actually it can only be done through the `_renderLabel` prop function and it's a bit complicated for just customizing the appearance of the default `<Text>`.


satya164:
Setting the focused style this way is not recommended since it doesn't play well with animations. You don't have to use `renderLabel` to customise the appearance, you can interpolate on `position` and pass the style as `tabStyle`. Please check the examples on how to do that.

PR #337, opened 2017-08-28
added useNativeDriver to animated library

erfanyousefifar:
i simply added useNativeDriver for making the process a little more smooth, it doesn't the tab touchable better but swiping in a noisy bridge mode is good and smooth.


satya164:
This won't work due to various reasons. Have you tested this?

PR #340, opened 2017-09-03
typescript declaration files

kaoDev:
Hey,
in a recent TypeScript project I used your lib and wanted to contribute my declarations

Greetings 
Kalle

### DIFF
--- a/package.json
+++ b/package.json
@@ -3,9 +3,8 @@
   "version": "0.0.69",
   "description": "Tab view component for React Native",
   "main": "src/index.js",
-  "files": [
-    "src/"
-  ],
+  "types": "src/index.d.ts",
+  "files": ["src/"],
   "scripts": {
     "test": "jest",
     "flow": "flow",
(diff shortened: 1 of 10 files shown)


wilomgfx:
Should probably be a @types/package.

kaoDev:
I think it is better to have the type-declarations direct in the package repository, but if this PR is not merged in the next days I will open a new PR in the @types repository

satya164:
@iRoachie can you review this PR?

iRoachie:
@satya164 I could, but I really advise that this be moved to DefinitelyTyped. DefinitelyTyped has tests that run against the typings, by using example use code. Right now there's no way for me to test if these typings work other than creating a new react-native project and using them.

Other than, the typings could actually all be done in one file, instead of 10 since your project imports come from one object.

If something upgrades in your lib, the process of rolling out that upgrade to the typings is a faster iteration over there as well. Also if over here, then you'll have to maintain them as well. The choice is up to you though.

kaoDev:
I have opened a PR in the definitely typed repo and will close this on
https://github.com/DefinitelyTyped/DefinitelyTyped/pull/20575

PR #346, opened 2017-09-17
Add PropTypes in TabBar to run custom focused tab

Kottidev:
I had problem WITH RTL and _renderIndicator when TabBar is scrollable 
so i added PropTypes to focusedtab **focusedLabelStyle** and **focusedTabContainerStyle**

image


satya164:
It's recommended to interpolate on the position value to customize focus styles as it wouldn't play nice with animations. I cannot add any props into core which doesn't play nice with the animations.

Kottidev:
@satya164  we need to fix the animations when the ScrollView is scrollable and app is isRTL.

with this case _renderIndicator is wrong

satya164:
Pull requests are welcome.

PR #348, opened 2017-09-21
Fix the screen flash when scrollEnable of tabbar is true

jiangleo:
The screen scrolls to the final position without animated, and the screen scrolls from the start position to the final position with animated. It causes the screen flash.

Because `componentDidUpdate` cause the screen to the final position, I deleted the `componentDidUpdate`, solved the problem.


satya164:

PR #349, opened 2017-09-21
fix : the position of tabbar is wrong when scrolling the pager manually.

jiangleo:
Before：

After:

`scrollAmount` is set by `_resetScrollOffset` and `_adjustScroll`, but only need `_resetScrollOffset`.


satya164:
Can you revert unrelated formatting changes and rebase against master? Also how did you repro this?

PR #355, opened 2017-09-26
Add accessibilityLable to TabBar.js

momoDeako:
I added getaccessibilityLabel to TabBar so that anyone who is going to use this library can pass accessibilityLabel and it will show up in the tab. Before I added the change, TabBar will take the accessibilityLabel from route, if route does not have any accessibility label, it will then use the title. 

Some background information: 
In some project, somehow route will set accessibility label as same as title.


momoDeako:
Will change the spacing soon.

satya164:
Closing due to inactivity.

PR #367, opened 2017-10-18
dismiss keyboard only when current position starts to differ from cur…

bunea:
…rent index

ViewPagerAndroid's "on-drag" seems to trigger keyboard dismissal even when it didn't scroll. It seems to receive an initial onPageScroll event (either after it measured the layout or when it mounted) with and initial position of 0 and offset 0. This causes any focused TextInput, either on the initial tab or on a modal, with autoFocus true to lose it's keyboard.

This pull request solves the issue by comparing the current index to the position of the view pager and hiding the keyboard only when they start diverging (e.g. sliding to another page)


satya164:

PR #376, opened 2017-10-31
BREAKING: add support for react-native-gesture-handler

satya164:
This adds a pager component based on [react-native-gesture-handler](kmagiera/react-native-gesture-handler), exposed as `TabViewPagerExperimental`. It'll allow you to use native animations and gestures to achieve much better performance.

This is a breaking change because the now pagers receive two new animated values (`panX` and `offsetX`) to control the `position` instead of a single `position` prop, and the `position` prop in no longer an `Animated.Value`. While you can still interpolate and do other operations on it, you cannot add listeners anymore or call `.setValue` on it. If you need to listen to updates, use the `position` event.

The other pagers have gone through a significant refactor to be able to co-exist with `TabViewPagerExperimental`. While the API stays the same, this can mean unexpected regressions. It'll be really helpful if you can give the different pagers a try and let me know if you encounter any issues.

You can try the demo here: https://exp.host/@satya164/react-native-tab-view-demos

### DIFF
--- a/example/package.json
+++ b/example/package.json
@@ -12,7 +12,7 @@
     "expo": "~22.0.0",
     "prop-types": "^15.5.10",
     "react": "16.0.0-beta.5",
-    "react-native": "~0.49.3"
+    "react-native": "https://github.com/expo/react-native/archive/sdk-22.0.1.tar.gz"
   },
   "devDependencies": {
     "babel-plugin-module-resolver": "^2.7.1",
--- a/.eslintrc
+++ b/.eslintrc
@@ -1,9 +1,10 @@
 {
   "extends": "eslint-config-satya164",
 
+  "plugins": ["react-native-globals"],
+
   "env": {
-    "browser": true,
-    "node": true,
-    "es6": true
+    "es6": true,
+    "react-native-globals/all": true,
   }
 }
--- a/.flowconfig
+++ b/.flowconfig
@@ -30,6 +30,8 @@
 .*/example/node_modules/reqwest
 .*/example/\.buckd/
 
+.*/node_modules/react-native-gesture-handler
+
 [include]
 
 [libs]
@@ -46,6 +48,7 @@ experimental.strict_type_args=true
 munge_underscores=true
 
 module.name_mapper='^expo$' -> 'emptyObject'
+module.name_mapper='^react-native-gesture-handler$' -> 'emptyObject'
 
 module.name_mapper='^[./a-zA-Z0-9$_-]+\.\(bmp\|gif\|jpg\|jpeg\|png\|psd\|svg\|webp\|m4v\|mov\|mp4\|mpeg\|mpg\|webm\|aac\|aiff\|caf\|m4a\|mp3\|wav\|html\|pdf\)$' -> 'RelativeImageStub'
 
--- a/.gitignore
+++ b/.gitignore
@@ -8,6 +8,7 @@
 # VSCode
 .vscode/
 tsconfig.json
+jsconfig.json
 
 # Xcode
 #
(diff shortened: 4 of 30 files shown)


RyszardRzepa:
It works really nice on Iphone 5 and Samsung Galaxy 6 edge.

PavSidhu:
Works great on iPhone 6S and Google Pixel XL

kkemple:
We're using this in the new MLS app, I'll test it out this week and get you some feedback!

hhg2288:
Swipe works perfectly in iPhone 6s+ :)

swipping works great, but clicking in tabs is a little janky sometimes. is this intended or is something wrong?

example/package.json
```diff
@@ -12,7 +12,7 @@
     "expo": "~22.0.0",
     "prop-types": "^15.5.10",
     "react": "16.0.0-beta.5",
-    "react-native": "~0.49.3"
+    "react-native": "https://github.com/expo/react-native/archive/sdk-22.0.1.tar.gz"
```
hhg2288: is this changed to support `ListView` ? because for `react-native@0.49` is Deprecated right?

kraffslol:
"7. With native animations" feels laggy in comparison with the others. Using iPhone 7.

Chathula:
It works well on iPhone 5. I will check this on android too.

brunolemos:
Worked pretty smooth on iPhone 6S. Some thoughts: 

- The demo could have the older implementation too, so we can compare it. 

- Maybe the pager should consider page change after 50% of the next one is visible

kelset:
Tested on Android 7.0.0 with a Samsung Galaxy S8 - works smoothly 👌

Ashoat:
This change breaks Flow types for anybody using this package, or using a package that depends on this package. Everybody using this package as a dependency will need to add an ignore to their .flowconfig for react-native-gesture-handler.

I think we should first address the broken flow types in react-native-gesture-handler. They should either be stripped entirely, or they should be fixed (preferable).

spencercarli:
When testing on React Navigation (particularly the [SimpleTabs](https://github.com/react-community/react-navigation/blob/master/examples/NavigationPlayground/js/SimpleTabs.js) example I notice a minor delay when switching tabs and the new active index styles to take hold.

This is on an iPhone 6.

example/package.json
satya164: @hhg2288 this expo version is equivalent to rn `0.49.3`. this change is not related to listview, migrating to flatlist can come in a future commit.

satya164:
> @spencercarli: minor delay when switching tabs and the new active index styles to take hold

Can you elaborate what these active index styles are? Are you talking about active tint color etc.?

> @Ashoat: This change breaks Flow types for anybody using this package

I get it, but that's not something I can take care of. someone who has a bit of more time than me has to send a PR to `react-native-gesture-handler`.

Thanks everyone for testing. I'm merging this to master and will soon publish a beta to npm. We can release the stable version when issues with flowtype are addressed in gesture handler.

Regarding the lag in iPhone 7, cc @kmagiera, did you experience any such issues or have an idea why this would happen? I don't have an iPhone 7 to test on. (https://twitter.com/roshangautam0/status/927466563875192832)

satya164:
New version `v0.0.71-beta.0` is on NPM with the `next` tag!

Ashoat:
https://github.com/kmagiera/react-native-gesture-handler/pull/52

kelset:
A-god stikes again 👏

spencercarli:
@satya164 I think you can actually view the bug on the example app you put together. If you go to the no animation example and click a new tab, the icon doesn't change. I'm guessing that on React Navigation we're going through an extra render cycle thus making it just seem like a delay.

Here's a [demo video](https://cl.ly/1d1p1T0q1V3D).

Issue #354, opened 2017-09-25
TabBar not Scrolling to proper position


yoginair:
tab bar not scrolling to proper position ,the tab indicator misplaced the position every-time i scrolled 
### Expected behaviour
![screen shot 2017-09-25 at 7 50 16 pm](https://user-images.githubusercontent.com/11831454/30813444-1571a1da-a22b-11e7-906d-edcd713d3273.png)
### Actual behaviour
![screen shot 2017-09-25 at 7 52 55 pm](https://user-images.githubusercontent.com/11831454/30813478-29616ac2-a22b-11e7-91fc-8e8bf68a71bf.png)
### Code sample, screenshots
 <TabBar
                {...props}
                scrollEnabled={true}
                indicatorStyle={{width:80,backgroundColor: "#515893"}}
                style={{backgroundColor: 'white',justifyContent:"center"}}
                tabStyle={{width:200}}
                labelStyle={{width:200,fontSize:20,color:"#515893",fontFamily:appFontFamily.APP_FONT_BOLD}}
            />

satya164:
Hey, I just released a new alpha `2.0.0-alpha.0` of the library. It's rewritten using [react-native-gesture-handler](https://github.com/kmagiera/react-native-gesture-handler) and [react-native-reanimated](https://github.com/kmagiera/react-native-reanimated) addresses a many platform specific bugs and performance problems. The documentation is updated as well.

Please try the new version and see if it addresses your issue. If not, please open a new issue following the issue template.

bookjan:
Hi, I share an example of the indicator width wrap the content width.

![Aug-08-2019 11-42-07](https://user-images.githubusercontent.com/1734406/62674985-70051800-b9d7-11e9-92c7-141b46423d68.gif)

The example code you can get it from [here](https://gist.github.com/bookjan/412d88ebd6b98f8ed49900be946668bf)

PR #378, opened 2017-11-01
Fix Tabbar scrolling wrong position

hieuxit:
The issue relate to this fix:
https://github.com/react-native-community/react-native-tab-view/issues/354


satya164:
should be fixed by f438b9900a81823799d2613864caf529363379ae

PR #379, opened 2017-11-02
Add activeLabelStyle prop

Symous:
Make us can set the style of the selected tab label text style. :D


satya164:
The `scene` object already includes a `focused` property. Also setting the active style this way is not recommended since it doesn't play well with animations. Please check the examples on how to interpolate on position to set tab style.

PR #383, opened 2017-11-04
Adds the option to use a custom style with the TabBar componen

lucasfernandes:
I noticed the need to use a custom style for this component, currently it is not possible to use an external style for this component, my changes are really simple and should not change the component behavior, since when a custom style is not defined the default style will be used instead.


satya164:
I cannot merge this pull request as a stylesheet is an implementation detail of the tab bar. Allowing to pass arbitrary stylesheets means exposing internal details as the library API. It prevents me from changing the structure of the tab bar without making it a breaking change.

Currently `TabBar` accepts various style props to customize different parts. If you need more control, please copy the tab bar code to your project and make necessary changes.

PR #394, opened 2017-11-16
Update TabViewPagerAndroid with safety check on _resetListener

mujavidb:
Similar to issue #365 , when react-native throws an error, this error is thrown instead as `undefined is not an object (evaluating 'this._resetListener.remove')`. 

This makes debugging the original problem much harder.

Not sure why practically any errors in react-native causes the `_resetListener` to be undefined, but in the mean time, this solution stops the problem.


mujavidb:
Is there a policy for when PRs are merged in?

PR #398, opened 2017-11-22
Added a upperCase property

mahomahoxd:
Instead of always having uppercase titles, by default there is no uppercase only if you pass the uppercase property.

### DIFF
--- a/src/TabBar.js
+++ b/src/TabBar.js
@@ -6,6 +6,7 @@ import {
   Animated,
   StyleSheet,
   View,
+  Text,
   ScrollView,
   Platform,
   I18nManager,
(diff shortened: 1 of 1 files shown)


satya164:
You can use the `getLabel` prop like `getLabel = ({ route }) => route.title`

mahomahoxd:
There is no such thing as getLabel in the properties that I have. I don't see any kind of properties or a way to set it without the uppercase (Without the use of a custom component of some sort). If you find me an alternative, sure. Otherwise I don't see a reason for not putting it in such a way like I already did.

satya164:
https://github.com/react-native-community/react-native-tab-view#tabbar-

wahidin32:
thats {{ route }} means, point to the array from the tab?

state = {
    index: 0,
    routes: [
      //key & title dari tabnya...
      { key: '1', title: '1' },
      { key: '2', title: '2' },
      { key: '3', title: '3' },
      { key: '4', title: '4' },
      { key: '5', title: '5' },
    ],
  };

PR #407, opened 2017-12-14
added bounces as passable prop to TabViewPagerScroll

ejkkan:


src/TabViewPagerScroll.js
```diff
@@ -134,7 +134,7 @@ export default class TabViewPagerScroll<T: *> extends React.Component<
         overScrollMode="never"
         scrollEnabled={this.props.swipeEnabled}
         automaticallyAdjustContentInsets={false}
-        bounces={false}
+        bounces={bounces ? bounces : false}
```
DaKaZ: I would cast to boolean instead of using a ternary here:
```
bounces={!!bounces}
```

satya164:
Closing due to inactivity

Issue #417, opened 2017-12-28
FEATURE: Vertical Tabs / Calculating Position


DaKaZ:
### Overview

This is not a bug, but a feature request (that I am also working on) for vertical tabs. I am having a small issue in the implementation and hoping for some help.

As a user of react-native-tab-view both natively and with React Navigation, it would be useful to be able to render a vertical tab bar especially for tablet landscape views.  I have begun and mostly completed a branch that implements this feature for TabViewAnimated.  

I have run into a problem and I am hopeful someone from the core team can help me.  I am implementing the animations for the tab selector now and, unfortunately, my lack of experience with animating is tripping me up. I am trying to follow a similar design pattern to the existing code, but I cannot figure out where `position` gets updated with the currently selected tab index.  I only see position set in the constructor of TabViewAnimated.  I did instrument the code (through a series of console.logs and the private `__getValue()` method) and I can see Position is updated, but how?  I just need to better understand this section of the code so that I can implement it for the vertical view.

### Expected behaviour

When I pass in the `vertical` prop to TabViewAnnimated
```
      <TabViewAnimated
        style={[styles.container, this.props.style]}
        navigationState={this.state}
        renderScene={this._renderScene}
        renderHeader={this._renderHeader}
        onIndexChange={this._handleIndexChange}
        initialLayout={initialLayout}
        vertical
      />
```
I expect to see:

![simulator screen shot - ipad air 2 - 2017-12-28 at 11 54 15](https://user-images.githubusercontent.com/895892/34424121-627810d4-ebde-11e7-843c-d718b9e2c5ab.png)

### Actual behaviour

The code does not current provide vertical tabs

satya164:
> where position gets updated with the currently selected tab index

`position` is a combination of 2 different animated values, `panX` and `offsetX`. Those values get updated inside the pagers, e.g.: https://github.com/react-native-community/react-native-tab-view/blob/master/src/TabViewPagerExperimental.js

DaKaZ:
@satya164 thank you!  I guess I was a little confused - aren't the pager files just for animating the content section?

satya164:
@DaKaZ pagers manage the gestures and the x-translation of the pages. the `position` animated value represents the current state of the x-translation.

DaKaZ:
[...]

PR #418, opened 2017-12-29
Vertical Tab Bar

DaKaZ:
This PR adds the ability render the tab bar vertically on the left with all the same features as the existing tab bar.  Fixes issue #417 

Currently, the implementation is only done for iOS (TabViewPagerScroll) but will add the other pager modes after an initial review.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,12 +1,9 @@
 {
   "name": "react-native-tab-view",
-  "version": "0.0.74-1",
+  "version": "0.0.74-2",
   "description": "Tab view component for React Native",
   "main": "src/index.js",
-  "files": [
-    "src/",
-    "types.js"
-  ],
+  "files": ["src/", "types.js"],
   "scripts": {
     "test": "jest",
     "flow": "flow",
@@ -25,14 +22,18 @@
   ],
   "repository": {
     "type": "git",
-    "url": "git+https://github.com/react-native-community/react-native-tab-view.git"
+    "url":
+      "git+https://github.com/react-native-community/react-native-tab-view.git"
   },
-  "author": "Satyajit Sahoo <satyajit.happy@gmail.com> (https://github.com/satya164/)",
+  "author":
+    "Satyajit Sahoo <satyajit.happy@gmail.com> (https://github.com/satya164/)",
   "license": "MIT",
   "bugs": {
-    "url": "https://github.com/react-native-community/react-native-tab-view/issues"
+    "url":
+      "https://github.com/react-native-community/react-native-tab-view/issues"
   },
-  "homepage": "https://github.com/react-native-community/react-native-tab-view#readme",
+  "homepage":
+    "https://github.com/react-native-community/react-native-tab-view#readme",
   "dependencies": {
     "prop-types": "^15.6.0"
   },
@@ -60,14 +61,8 @@
   },
   "jest": {
     "preset": "react-native",
-    "setupFiles": [
-      "<rootDir>/__setup__/enzyme.js"
-    ],
-    "modulePathIgnorePatterns": [
-      "<rootDir>/example/node_modules"
-    ],
-    "snapshotSerializers": [
-      "enzyme-to-json/serializer"
-    ]
+    "setupFiles": ["<rootDir>/__setup__/enzyme.js"],
+    "modulePathIgnorePatterns": ["<rootDir>/example/node_modules"],
+    "snapshotSerializers": ["enzyme-to-json/serializer"]
   }
 }
--- a/src/TabViewPagerPan.js
+++ b/src/TabViewPagerPan.js
@@ -259,11 +259,15 @@ export default class TabViewPagerPan<T: *> extends React.Component<Props<T>> {
           <View
             key={navigationState.routes[i].key}
             testID={navigationState.routes[i].testID}
-            style={
+            style={[
               width
                 ? { width }
-                : i === navigationState.index ? StyleSheet.absoluteFill : null
-            }
+                : i === navigationState.index ? StyleSheet.absoluteFill : null,
+              // prevent keyboard navigation to off-screen pages
+              i !== navigationState.index && Platform.OS === 'web'
+                ? { visibility: 'hidden' }
+                : null,
+            ]}
           >
             {i === navigationState.index || width ? child : null}
           </View>


satya164:
Thanks for the pull request, but I don't really want to maintain this feature in the repo, especially in the tab bar which took a significant effort to make it scrollable. I'm the only person who works on the library and it's already a huge undertaking. I'd be interested in a PR which does minimal changes to the pagers to support vertical pagination and the tabbar is kept separate repo instead of changing the current tabbar, but I am not really sure that this is the correct approach. Any landscape mode apps I have used or very different from what this library does.

There are some fundamental issues that need to be addressed:
- Normally each page in a tab contains a scrollview. When you have scrollviews inside each page, how would you swipe vertically?
- I haven't used any apps where it animates the pages vertically when you change pages, it doesn't make sense to me because it will conflict with the scroll
- The sidebar (or as you call it, tabs) is usually a simple scrollable list in the apps I used and doesn't need to synchronise with the animation position of the right pane (at least I have never seen such a thing)

DaKaZ:
@satya164 Thanks for your comments.  Allow me to address a few:

> I'm the only person who works on the library and it's already a huge undertaking
I certainly have a better appreciation for the complexity of this library but thankfully I am much more familiar with all the code now.  If you ever need help please let me know.

> the tabbar is kept separate repo
Did you really mean another repo all together?  I definitely understand your concerns over maintainability, that is why I worked hard to maintain the same design patterns and make minimal changes.  If you do look at the change set, you'll see that there are really only a few additional variables/props that allow us to maintain the vertical tabs.

> Any landscape mode apps I have used or very different from what this library does
I mostly agree - the current PR was an attempt to make sure the vertical approach was feasible and would be accepted by the library's maintainers :) . I feel confident that I could address the scrollability (ie: make the pagers still a horizontal scroll while the tabbar is vertical.

I definitely do not want to put undo burden on you or this library, but I do feel like this is a feature that many people are looking for.  I'd like to find a way to integrate it in a maintainable way.  Let me know your thoughts.

DaKaZ:
@satya164 I just pushed a new commit that makes leaves the pager as horizontal scroll just like a regular tab bar.  Overall I think this is much better functionality AND its actually a bit cleaner code too.  Please take a look and let me know your thoughts.

satya164:
> Did you really mean another repo all together

Yes

> If you do look at the change set, you'll see that there are really only a few additional variables/props that allow us to maintain the vertical tabs

There are still significant additions in the tab bar. And I'm not convinced that the vertical tab bar should behave the same way as the horizontal tab bar. It should behave like a sidebar. And it should be a separate component because of the different behaviour. Also I'm don't really know the behaviour you're trying to emulate. Can you post a GIF and link to an app/mockup that demonstrates this design pattern?

I want it in a separate repo because I won't be using this feature in near future, and it's better you maintain it if you're using the feature.

I would accept changes to the pager, which will enable the you to build vertical tab bar using the same pager, but I don't want to modify the current tab bar component. But again, I need to see this design pattern first.

>  I feel confident that I could address the scrollability (ie: make the pagers still a horizontal scroll while the tabbar is vertical

It's very uncommon to have horizontally scrollable areas (except paginated ones and carousels). A common design pattern is to have vertically scrollable areas, which won't work with vertical pagers.

So I basically need this from you:
- Post a GIF of the pattern you are building currently so I have a better understanding
- Link to some places where this design pattern is used, could be an app, a mockup, anything, just need to see someone using this pattern

If we establish that this is a common pattern, we need to:
- Change all of the pagers to accept a `vertical` prop
- Rename `offsetX` and `panX` animated values to `offset` and `pan` to make them generic
- Switch to `flexDirection: row` in `TabViewAnimated` when `vertical === true`

Regarding your current code:
- You should reuse the `layout` prop instead of adding a `onLayout` inside the pager, I don't understand why the height measured by `TabViewAnimated` would be different from the one measured by `TabViewPagerScroll`?
- There are 4 different pagers. The functionality needs to be implemented in all of the ones that support it, i.e. `PagerPan`, `PagerExperimental` for both platforms and `TabViewPagerScroll` for iOS

DaKaZ:
@satya164 thank you again for the very detailed response - I really appreciate the time you are taking with this.

> I want it in a separate repo because I won't be using this feature in near future, and it's better you maintain it if you're using the feature.

While this is unfortunate, I completely understand.  Lets finish our discussion and if we decide that we should indeed maintain this separately I'll be glad to open another repo.

> You should reuse the layout prop instead of adding a onLayout inside the pager, I don't understand why the height measured by TabViewAnimated would be different from the one measured by TabViewPagerScroll?

I don't think that will work.  The problem we are solving here is not height, but width of the content area in the pager.  When the tabs are added vertically, we are consuming some of the initial layout space of the pager's content area.  In order for us to properly calculate the index and animate the active indicator to the correct location, we need the offsetX value divided by the width.  Thus, we need to know the final layout WIDTH of the pager area and pass this back up to TabViewAnimated.  I don't know how else to query an elements final layout from the parent.  This was the bug I mentioned earlier where the active indicator was not tracking properly, it took me a while to figure it out :)

> There are 4 different pagers. The functionality needs to be implemented in all of the ones that support it, i.e. PagerPan, PagerExperimental for both platforms and TabViewPagerScroll for iOS

Of course, as I mentioned before I was only holding off on implementing those until we had consensus on the approach.

> Rename offsetX and panX animated values to offset and pan to make them generic

This is not needed now as the pager only animates in the X direction regardless of the tab's orientation.

> Switch to flexDirection: row in TabViewAnimated when vertical === true

This is already done on line 214: https://github.com/react-native-community/react-native-tab-view/pull/418/files#diff-27268ea1510a16d194c3d7b78dfa0e6dR214

> Post a GIF of the pattern you are building currently so I have a better understanding

verticaltabview

> Link to some places where this design pattern is used, could be an app, a mockup, anything, just need to see someone using this pattern

I am working on this now, I don't have a lot of 3rd party examples, but in our use case when we switch from phone to tablet/landscape we are wanting the tabs to go from the bottom to the left vertically.  For me, we already have all our code and routes laid out, so re-implementing this in a "custom" sidebar would be crazy duplicative.  In order to keep the code dry, I wanted to reuse the same tabbar code we already have in the app.  That is why I went this direction.  Here is our designer's screen of the tablet landscape view for our app:

verticaltabdesign

DaKaZ:
@satya164 I have added the changes to all the other pagers and I also included some additional tests to help verify the functionality.  I could not figure out how to mock the `onLayout` event for Android's ViewPagerAndroid so if you have any ideas, that would be great.

satya164:
> I don't think that will work. The problem we are solving here is not height, but width of the content area in the pager.

Looks like wrapping the `renderPager(...)` call in a `View` and moving the `onLayout` call there would solve the issues you have with layout and will not require separate `onLayout` on the pagers?

https://github.com/react-native-community/react-native-tab-view/blob/master/src/TabViewAnimated.js#L195

Also looking at the GIF, that should be the only change required to address your use case. I don't even need a `vertical` prop since you should be able to pass `style={{ flexDirection: 'row' }}` to `TabViewAnimated`.

DaKaZ:
> Looks like wrapping the renderPager(...) call in a View and moving the onLayout call there would solve the issues you have with layout and will not require separate onLayout on the pagers, and will also keep the code simple.

Brilliant!  Will look at this in the morning as well as address your other comments.

DaKaZ:
@satya164 I implemented your genius idea of wrapping the pager in a view and that works GREAT!  There are no changes needed now to any of the pagers!  I rebased and squashed the code... there is a full working example in the examples directory.  I actually think that the changes are so minor you might consider accepting this PR :)

DaKaZ:
@satya164 Sorry, I lied - there are a few updates needed in the pagers.  Its incredibly small, but without them the position is off when you switch to/from vertical - like when orientation changes :) . I also fixed a small bug where the active indicator wasn't rendering correctly on Android.

satya164:
> but without them the position is off when you switch to/from vertical - like when orientation changes

I don't think checking the `vertical` prop in pagers is the correct way. The pagers aren't paged vertically, so the prop is confusing. When you are changing the `vertical` prop, what's actually changing is the layout. So the pagers need to respond to layout changed properly and whatever you're doing on `vertical` prop change could be done on `layout` prop change instead.

DaKaZ:
> The pagers aren't paged vertically, so the prop is confusing. 

I agree with that statement.

> When you are changing the vertical prop, what's actually changing is the layout. 

This is true as well.

However.... some of the pagers currently have a bit of a "short circuit" that we'd need to remove.  For example:
https://github.com/react-native-community/react-native-tab-view/blob/master/src/TabViewPagerAndroid.js#L69

Do you think its OK to remove this or would suggest a different approach?

lucasleelz:
I like this

Issue #413, opened 2017-12-21
Detected division by zero in AnimatedDivision


quy1403:
When I use ScrollableTabView in Tab screen by wix react-native-navigation, I got the red screen with error "Detected division by zero in AnimatedDivision"
### Expected behaviour
work well
### Actual behaviour
 I got the red screen with error "Detected division by zero in AnimatedDivision"
### Code sample, screenshots
  state = {
        index: 0,
        routes: [
            {key: '1', title: 'TAB A'},
            {key: '2', title: 'TAB B'},
        ],
    };

    _handleIndexChange = index =>
        this.setState({
            index,
        });

    _renderBadge = ({route}) => {
        if (route.key === '2') {
            return (
                <Badge>{12}</Badge>
            );
        }
        return null;
    };
    _renderLabel = props => ({route, index}) => {
        const inputRange = props.navigationState.routes.map((x, i) => i);
        const outputRange = inputRange.map(
            inputIndex => (inputIndex === index ? COLORS.MAIN_COLOR : '#222')
        );
        const color = props.position.interpolate({
            inputRange,
            outputRange,
        });

        return (
            <Animated.Text style={[styles.label, {color}]}>
                {route.title}
            </Animated.Text>
        );
    };
    _renderHeader = props => (
        <TabBar
            {...props}
            scrollEnabled
            indicatorStyle={styles.indicator}
            style={styles.tabbar}
            tabStyle={styles.tab}
            renderLabel={this._renderLabel(props)}
            renderBadge={this._renderBadge}
            pressOpacity={1}
        />
    );

    _renderScene = ({route}) => {
        switch (route.key) {
            case '1':
                return (
                    <FeeTab
                        navigator={this.props.navigator}
                    />
                );
            case '2':
                return (
                    <FeeTab
                        navigator={this.props.navigator}
                    />
                );
            default:
                return null;
        }
    };

    renderPager(props) {
        return <TabViewPagerPan {...props} />;
    }

    render() {
        return (
            <PageTabBar>
                <StatusBar />
                <TabViewAnimated
                    style={[styles.container, this.props.style]}
                    navigationState={this.state}
                    renderScene={this._renderScene}
                     renderHeader={this._renderHeader}
                    onIndexChange={this._handleIndexChange}
                    initialLayout={{
                        height: 0,
                        width: Dimensions.get('window').width,
                    }}
                    renderPager={this.renderPager}
                    useNativeDriver
                />
            </PageTabBar>

[...]

PR #442, opened 2018-02-02
[FIX] Animated devided by zero

mckn:
We are using this library with `wix/react-native-navigation` and have some issues with the error described in issue #413 .

This PR prevents the application from crashing if one of `height` or `width` cant be measured for some reason during the onLayout lifecycle.

### DIFF
--- a/src/TabViewAnimated.js
+++ b/src/TabViewAnimated.js
@@ -132,6 +132,10 @@ export default class TabViewAnimated<T: *> extends React.Component<
       return;
     }
 
+    if (!height || !width) {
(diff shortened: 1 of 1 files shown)


src/TabViewAnimated.js
```diff
@@ -132,6 +132,10 @@ export default class TabViewAnimated<T: *> extends React.Component<
       return;
     }
 
+    if (!height || !width) {
```
satya164: This should not happen since we have a check below at L142. Can you provide a repro?

PR #450, opened 2018-02-12
When used with react-native-web, do not expect ViewPagerAndroid to be supported

raarts:
react-native-web aliases `import ... from 'react-native'` into `import ... from 'react-native-web'`.
But it does not implement undocumented or iOS/Android-only components, not even stubs for it.
So I added a stub in this package, which provides a component which just says: not implemented.


raarts:
@satya164 Could you please take a look at this?

PR #460, opened 2018-02-22
Add onSwipeStart and onSwipeEnd props to PagerScroll and PagerAndroid

charpeni:
These props were already available in `TabViewPagerPan`, but they would also be interesting in `TabViewPagerScroll` and `TabViewPagerAndroid`.

I plan to use this in react-navigation to attach and detach sibling tabs accordingly to swipe events.


satya164:
> I plan to use this in react-navigation to attach and detach sibling tabs accordingly to swipe events.

What do you mean by attach/detach siblings? For lazy loading?

charpeni:
Exactly. 

I've improved Brent's `ResourceSavingSceneView` component, this PR and the next react-native-tab-view release will allow that feature to land in react-navigation.

Here's a sneak peek. :tada:

#### Lazy loaded on swipe and unclipped once they're out of the screen. On the transition from 1 to 5, 2-4 are unclipped.
react-navigation-fullprops

#### Lazy loaded on focus
react-navigation-lazyonfocus

#### Lazy loaded on focus, but all screens stay visible, even while transitioning.
react-navigation-alwaysvisible

PR #475, opened 2018-03-05
Use internal current index in TabViewPagerPan instead of the one from props

### LINKED ISSUE #448: Strange behavior (overswipe) in Coverflow example
Hi! This library is great, thanks for maintaining and improving it!

I was checking `Coverflow` example from [React Native Tab View Demos](https://expo.io/@satya164/react-native-tab-view-demos) on expo and noticed strange behavior.

### Expected behaviour
I expected the first and the last slides either to disallow further swipe at all, or to become smaller/semi-transparent and on touch end return to normal.

### Actual behaviour
I guess as long as spring animation hasn't fired "on completed" callback (I'm not sure if I understand the concept of it) or something, the first and the last slides allow overswipe with behavior, that doesn't seem intended.
Last slide reacted to overswipe by becoming larger instead of smaller.
First slide reacted to overswipe by moving in the opposite of expected direction.

### Code sample, screenshots
See [the video](https://youtu.be/AgOisy7Vy1c), pay attention to touch indicator. With the last slide spring animation ended after the first overswipe and hasn't reacted to it afterwards, but with the first slide with quick swipes I was able to trigger a lot of strange animations.

### What have you tried
To swipe here and there.

zamotany:
Index from `props` might be out of sync due to async nature of `setState`.

### Motivation

Fixes #448 

### Test plan

Manual testing due to specific nature of the problem.

### DIFF
--- a/example/src/CoverflowExample.js
+++ b/example/src/CoverflowExample.js
@@ -50,8 +50,7 @@ export default class CoverflowExample extends React.Component<*, State> {
     const { width } = layout;
     const { routes } = navigationState;
     const currentIndex = routes.indexOf(route);
-    // Prepend '-1', so there are always at least 2 items in inputRange
-    const inputRange = [-1, ...routes.map((x, i) => i)];
+    const inputRange = routes.map((x, i) => i);
     const translateOutputRange = inputRange.map(i => {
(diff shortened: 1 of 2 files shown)


satya164:
Thank you for your pull request!

PR #476, opened 2018-03-05
Add prop numberOfLines to TabBar

jonathanort:
### Motivation

Making sure the app looks good even with very long TabBar text. Useful for us with dynamic text in it.

### DIFF
--- a/src/TabBar.js
+++ b/src/TabBar.js
@@ -177,7 +177,7 @@ export default class TabBar<T: *> extends React.Component<Props<T>, State> {
       return null;
     }
     return (
-      <Animated.Text style={[styles.tabLabel, this.props.labelStyle]}>
+      <Animated.Text style={[styles.tabLabel, this.props.labelStyle]} numberOfLines={this.props.numberOfLines}>
         {label}
       </Animated.Text>
     );


src/TabBar.js
```diff
@@ -177,7 +177,7 @@ export default class TabBar<T: *> extends React.Component<Props<T>, State> {
       return null;
     }
     return (
-      <Animated.Text style={[styles.tabLabel, this.props.labelStyle]}>
+      <Animated.Text style={[styles.tabLabel, this.props.labelStyle]} numberOfLines={this.props.numberOfLines}>
```
jmitchell38488: A callback is provided to allow the dev to provide alternative label rendering, even if it's just copy & pasting the original code

satya164:
Thank you for the PR!

> Making sure the app looks good even with very long TabBar text

IMO if your tab label text is so long that it needs to be clipped, better to change it to be smaller or increase the tab width using the `tabStyle` prop. A tab label is a navigation item, and not being able to see the full label for where I'm navigating is a poor UX.

The default tab bar follows material design guidelines which recommend against this: https://material.io/guidelines/components/tabs.html#tabs-content

I won't be merging because of this. But you can use `renderLabel` to render your custom label which handles this.

PR #481, opened 2018-03-12
Added active label style prop in the top bar.

lpotapczuk:
### Motivation

As of today, there is no way to pass styles that indicate active label in TabBar. In my pull request, I've added new prop that enables user to specify styles of active label. To do this, you need to pass prop I've called `activeLabelStyle`. I've added this case to `TopBadTextExample.js` so that you can see how this works.

### Test plan

Pass `activeLabelStyle` prop to TabBar to indicate with style of an active tab (see screenshot below):

29103593_10157178962752802_1735794565773787136_n

### DIFF
--- a/example/src/TopBarTextExample.js
+++ b/example/src/TopBarTextExample.js
@@ -50,6 +50,7 @@ export default class TopBarTextExample extends React.Component<*, State> {
       style={styles.tabbar}
       tabStyle={styles.tab}
       labelStyle={styles.label}
+      activeLabelStyle={styles.activeLabel}
     />
   );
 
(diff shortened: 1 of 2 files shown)


satya164:
From https://github.com/react-native-community/react-native-tab-view/pull/379#issuecomment-343634037

A better approach to support `activeLabelStyle` and `inactiveLabelStyle` would be to overlay one on top of another and crossfade them. This would also make it easy to animate things like animating color.

I'm closing this since doing it this way will cause a jump in the style. Please send another PR with the cross fading if you want to work on it :)

Issue #483, opened 2018-03-14
Strange tab selection behaviour


wangbourne:
### Current behaviour
When I click one tab item, it behaviours like a spring: 

- if I swipe **left(right)** and then click one tab item, then it automatically scroll **left(right) couple pages**, then come back to the position where I select that tab item;

### Expected behaviour

Click one tab item and jump to that item directly

### Screenshots (if applicable)

1. swipe right first

![2](https://user-images.githubusercontent.com/2090084/37432101-ff37ffc6-2822-11e8-8562-0119f477fa88.gif)
<br>
- swipe left first

![3](https://user-images.githubusercontent.com/2090084/37432435-0a225408-2824-11e8-98d8-fdea79d2fd1a.gif)

### Your Environment

| software              | version
| --------------------- | -------
| ios or android        |   iOS10,11
| react-native          |  0.51.10
| react-native-tab-view |  0.0.74
| node                  |    v9.4.0
| npm or yarn           |    1.5.1 (Yarn)

satya164:
Do you have an RTL layout?

wangbourne:
I guess RTL is a term for Android, right? Actually, I'm working on iOS only

wangbourne:
Below is whole my code here.  One more finding here is: if I comment out the logic in `_onTabPress`, it would be like this: no spring back, but wrong selected position, i have to manually scroll tab to show selected bar.

![3](https://user-images.githubusercontent.com/2090084/37433807-19fcbbc0-2829-11e8-9ad2-8ca9ed808a44.gif)


```
import {findIndex as _findIndex} from 'lodash';
import * as React from 'react';
import { Dimensions, StyleProp, ViewStyle} from 'react-native';
import {
  NavigationState,
  Route,
  Scene,
  SceneMap,
  TabBar,
  TabBarProps,
  TabViewAnimated,
} from 'react-native-tab-view';
import styles from './Styles';

const initialLayout = {
  height: 0,
  width: Dimensions.get('window').width,
};

export type TabData = Route<{
  key: string;
  title: string;
  id: string;
}>;

type State = NavigationState<
  TabData
>;

export interface Props {
  handleIndexChange(route: TabData): void;
  renderScene(route: TabData): React.ReactNode;
  tabs: TabData[];
  initialPage: number;
  style?: StyleProp<ViewStyle>;
}

export class Tabs extends React.Component<Props, State> {
  constructor(props: any) {
    super(props);
    this.state = {
[...]

Issue #482, opened 2018-03-13
initialLayout={initialLayout} could slow down the application


wangbourne:
### Current behaviour
My application comes a performance issue. After trying, I found if I remove `initialLayout={initialLayout}`, my application become much faster. What could be the possible reason for this?

By the way, how can I use `TabViewPagerExperimental`?  It does not come in `@types/react-native-tab-view"`  (I'm using Typescript)

### Code sample

```
return (
      <TabViewAnimated
        style={[styles.container, this.props.style]}
        navigationState={this.state}
        renderScene={this._renderScene}
        renderHeader={this._renderHeader}
        onIndexChange={this._handleIndexChange}
        initialLayout={initialLayout}
      />
    );
```

```
_renderScene({ route }: any) {
   // Optimize large number of routes
    const routeIndex = _findIndex(this.state.routes, (item: any) => item.id === route.id);
    if (Math.abs(this.state.index -  routeIndex) > 2) {
      return null;
    }
    //console.log(routeIndex);

    return this.props.renderScene(route);
  }
```
here, `this.props.renderScene` will be called in parent component like

```
//Avoid unnecessary re-renders
renderScene = (route: TabData) => {

    const { sports, match, itemsWithKey, history, handleRouting } = this.props;

    const matches = [...itemsWithKey[route.id].match];

    return (<EventListView match={match} history={history} items={matches} onClick={this.handleEventListClick}/>);
  }
```

### Your Environment

| software              | version
| --------------------- | -------
| ios or android        |   iOS10,11
| react-native          |  0.51.10
| react-native-tab-view |  0.0.74
| node                  |    v9.4.0
| npm or yarn           |    1.5.1 (Yarn)

satya164:
> I found if I remove initialLayout={initialLayout}, my application become much faster. What could be the possible reason for this?

I don't know. What do you mean by faster? Are you talking about initial render time? In that case, I can see how it could make it a bit slower since all scenes are rendered at once when `initialLayout` is specified.

> how can I use TabViewPagerExperimental? It does not come in @types/react-native-tab-view"

I don't maintain the typings and I don't know who does it. You have to find the person who added them.

wangbourne:
As to your answer to `initialLayout={initialLayout}`, you have explained well and thanks :)

By the way, why each scene has to be called twice after render()? I can see following logs from chrome. (My scene rendering is expensive .)

[...]

Issue #424, opened 2018-01-18
Specifying non-0 index resets header immediately back to start.


browniefed:
### Expected behaviour
Should select specified index

### Actual behaviour
Renders at proper index, then rebounds to the start immediately.

### Code sample, screenshots
![tabissue](https://user-images.githubusercontent.com/1714673/35075451-eb158c4a-fba7-11e7-94d7-9a259d200d0a.gif)

Seemingly the issue is only the header, the rest of the body renders as intended. Later if you touch it, then it will reset.

```
  _renderHeader = props => (
    <TabBar
      {...props}
      tabStyle={{
        width: 75,
      }}
      scrollEnabled
      indicatorStyle={{ backgroundColor: "#FFF" }}
      style={{ backgroundColor: blue }}
    />
  );
```

Rest of the render is nothing special

```
 _renderScene = SceneMap({
    APPL: Route,
    TWTR: Route,
    TSLA: Route,
    NFLX: Route,
    FB: Route,
    MSFT: Route,
    DIS: Route,
  });
  render() {
    const state = this.state;

    return (
      <View style={styles.container}>
        <TabViewAnimated
          style={styles.container}
          navigationState={this.state}
          renderScene={this._renderScene}
          renderHeader={this._renderHeader}
          onIndexChange={this._handleIndexChange}
          initialLayout={initialLayout}
        />
      </View>
    );
  }
```

### What have you tried

Tried removing tabStyle width, removed scrollEnabled.

browniefed:
More info. Seems to only occur if remote JS debugging is enabled.

However now it seems to just send it to the end of the container when remote JS debugging is disabled.

![tabissue2](https://user-images.githubusercontent.com/1714673/35075635-178ef76a-fba9-11e7-81ee-8c9a05a9710c.gif)

dtuan:
@browniefed I'm using TabViewPagerScroll to "renderPager" and have this issue too, my quick fix is update **_handleScroll** function a little bit to:
```
_handleScroll = (e: ScrollEvent) => {
    if (this._isInitial || e.nativeEvent.contentSize.width === 0) {
      return;
    }
    ......
} 
```

EvianZhow:
Having this issue either, disable `Remote JS Debugging` and `Hot Reloading` works normally. Enable either option cause the index tab reset back to 0.

EvianZhow:
@browniefed @satya164 

[...]

PR #491, opened 2018-03-23
Couple bugs fix for many tabs 

wangbourne:
### Motivation
This PR is to resolve couple issues talked in followings, especially on the low performance devices like iPhone5

1. https://github.com/react-native-community/react-native-tab-view/issues/483
2. https://github.com/react-native-community/react-native-tab-view/issues/482  (Same as https://github.com/react-native-community/react-native-tab-view/issues/424)
3. if tab 1 is selected, then scroll tab bar directly to tab 20, then click tab 20, you can see uncomfortable animations

### DIFF
--- a/example/src/TopBarTextExample.js
+++ b/example/src/TopBarTextExample.js
@@ -33,31 +33,7 @@ export default class TopBarTextExample extends React.Component<*, State> {
       { key: 'article', title: 'Article' },
       { key: 'contacts', title: 'Contacts' },
       { key: 'albums', title: 'Albums' },
-      { key: 'chat1', title: 'Chat' },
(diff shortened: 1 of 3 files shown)


satya164:
Thanks for your PR! I merged it here with some minor tweaks: https://github.com/react-native-community/react-native-tab-view/commit/0c6aacccf0fd21c7537537acc4ef08a5ada349e0

Regarding disabling animations completely when jumping to tab far away, I reverted it for now since I'm not sure about it. Maybe jumping to nearest tab without animation and then jumping to the desired tab with animation provides a better UX. Please open a new PR with that if you want :)

PR #494, opened 2018-03-27
TabViewPagerPan compatible with SafeAreaView

bkDJ:
On iPhone X, content in a [SafeAreaView](https://facebook.github.io/react-native/docs/0.50/safeareaview.html) allows its children to avoid being under the home bar and sensor housing of iPhone X. This relies on all parent views also being bounded by the screen edges. That limitation is so that content can scroll offscreen without being clipped at the safe area edges.

In an app the team I'm on is building, we are using React-Navigation which uses React-Native-Tab-View's TabViewPagerPan for bottom-tabs on iOS. Currently, the parent of all the tabs is `number_of_tabs` times larger than the screen, and SafeAreaView is not working correctly. For example, an iPhone X in landscape would have the SafeAreaView padding in the first tab only present on the left. The final tab would be padded on the right. And all tabs between those would have no padding at all (components near the screen edge would be unreachable or cut off).

This Pull Request adjusts the parent to be the size of the screen, and its tab children are translated with the Pan gesture and tab switching. In the screenshots below, blue is where content will draw inside of a SafeAreaView.

||Left tab | Middle tab | Right Tab|
|---|---|---|---|
|*Before*||||
|*After*||||

### DIFF
--- a/src/TabViewPagerPan.js
+++ b/src/TabViewPagerPan.js
@@ -262,7 +262,7 @@ export default class TabViewPagerPan<T: *> extends React.Component<Props<T>> {
               width
                 ? {
                     width,
-                    transform: [{ translateX }]
+                    transform: [{ translateX }],
                   }
                 : i === navigationState.index ? StyleSheet.absoluteFill : null
             }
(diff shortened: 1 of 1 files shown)


Boris-c:
Hi there,

Is someone planning to review that PR? The fix proposed by bkDJ is awesome, and it'd be great to have it reviewed and pulled...

Thanks!

Boris

satya164:

PR #497, opened 2018-04-02
RTL support for TabViewPagerPan

farnabaz:
When use Tabbar in RTL apps, first tab shown perfect and as expected but other tabs are blank and doesn't show anything.
After digging in library I found that this issue is related to transformation of View inside `TabViewPagerPan.is`  and it could solve just by inverting translate value for RTL layouts

### Test plan
- Just add tabbar with two or three tabs and use `I18nManager.forceRTL(true)` to change layout direction


satya164:
Thank you for the PR!

PR #519, opened 2018-05-12
tabBar: Pass borderless from props to TouchableItem.

jainkuniya:
Before always true was passed.

### Motivation
We are using this and was facing issue of extended ripple effect. Because always true was passed. So pass it from the the props, giving control to the user.
37422059-98921d98-27e0-11e8-90d2-e23636c589b4

 

### Test plan

Pass `borderless: false,` or `borderless: true,` in the `tabBarOptions`.

### DIFF
--- a/src/TabBar.js
+++ b/src/TabBar.js
@@ -24,6 +24,7 @@ type IndicatorProps<T> = SceneRendererProps<T> & {
 };
 
 type Props<T> = SceneRendererProps<T> & {
+  borderless?: boolean,
   scrollEnabled?: boolean,
   bounces?: boolean,
   pressColor?: string,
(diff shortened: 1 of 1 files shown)


jainkuniya:
Updated!
Now checks are passing.

@satya164 please review it, when you get some time :)

satya164:

chrisbobbe:
I'm still seeing the issue (in the same project as @jainkuniya) with the rewritten version, react-native-tab-view 2.15.1. However, I don't see it in the react-native-tab-view example app. I could repeat this PR, but I'm thinking there may be pretty substantial changes with the new [`Pressable` API](https://reactnative.dev/blog/2020/07/06/version-0.63#pressable) with RN v0.63, and it might naturally be solved at that point. In the meantime, we have a workaround, setting `borderWidth` and `elevation` to zero (see zulip/zulip-mobile@29d1da6c28fb5e31426c3602acc186d6f9fe8e0e).

PR #520, opened 2018-05-14
Added `activeTabStyle` and `activeLabelStyle` option for `TabBar` component.

alimek:
### Motivation

As of today, there is no way to pass styles that indicate active label in TabBar and also for active tab bar item. I have added option to pass `activeTabStyle` and also `activeLabelStyle` (thanks for @	lpotapczuk - merged his PR with `activeLabelStyle` and fixed eslint to pass eslint and flow checks)

### Test plan

Pass `activeTabStyle` and `activeLabelStyle` to `TabBar` component in `_renderHeader` function to modify style on active element.

### DIFF
--- a/README.md
+++ b/README.md
@@ -148,6 +148,7 @@ renderTabBar={props =>
 - `tabStyle`: style object for the individual tabs in the tab bar.
 - `indicatorStyle`: style object for the active indicator.
 - `activeTabStyle` - style object for the individual active tab in the tab bar
+- `activeLabelStyle` - style object for the individual label in the active tab bar
 - `labelStyle`: style object for the tab item label.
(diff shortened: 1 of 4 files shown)


r0st1kuzh:
Could you merge this one?

kristerkari:
Would be nice to be able to set the active style for labels, as it seems to be missing from the library.

Is there some reason not to merge this @satya164?

satya164:
This PR is similar to #379 and #481

From my comment in #481

> A better approach to support `activeLabelStyle` and `inactiveLabelStyle` would be to overlay one on top of another and crossfade them. This would also make it easy to animate things like animating color.

Please send another PR if you want to implement it. Thanks :)

PR #528, opened 2018-05-24
feat: add the ability to override pager container style

agurtovoy:
### Motivation

One specific use case for this is implementing variable tab height support. Default
`pager: { flex: 1 }` style interferes with it; this change gives us a way to override it.

### Test plan

This is a minimal change that does not change current behavior of the component. It does introduce another point of customization for something that wasn't customizable, but since similar existing points of customization are not being automatically tested, I did not extend the test suite to cover this functionality.

### DIFF
--- a/README.md
+++ b/README.md
@@ -119,6 +119,7 @@ Container component responsible for managing tab transitions.
 - `renderFooter` - optional callback which returns a react element to use as bottom tab bar
 - `renderPager` - optional callback which returns a react element to handle swipe gesture and animation
 - `renderScene` - callback which returns a react element to use as a scene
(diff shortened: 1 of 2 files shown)


satya164:
> This is a minimal change that does not change current behavior of the component

It's an implementation detail which I don't want to expose. Can't you just pass style to the pager using `renderPager` to achieve what you need?

agurtovoy:
@satya164 You can't; the pager itself is wrapped into a container with a hard coded style, and that's what this change is about. Sorry for not being 100% clear.

satya164:

Issue #356, opened 2017-09-28
[RFC] Dynamic tab width


sospedra:
I'm planning to make a PR adding the desired and very intrinsic material design feature of having dynamic width sizes for each tab. By extension, other problems like multilines or ellipsis will be addresses to.

I made some proof of concepts and I find a proper way to make it dynamic. Tho, it relies onto `onLayout`. Before invest a hard time making the PR I'm wondering if you can see any downside with this approach or shall I continue with the proposal.

cc @satya164

sospedra:
😞

satya164:
Hey. Sorry I was busy. Feel free to open a WIP PR. I'll take a look.

Tim152:
@sospedra hey! I'am waiting for your PR too :)

lucasleelz:
@sospedra hey! I'am waiting for your PR too :)

anoopmm:
is there any updates

sospedra:
I was working on this the past week. Is much more complicated than I thought.
My idea was to wait until the tabs are rendered and the get the width from the layout.
Problem is with the total width of the tabbar which works terribly bad with this dynamic idea.
I still think is possible but a bigger refactor is needed beforehand.

If anyone is interested in this feature and is willing to work on that I'm open to chat about my research so far.

satya164:
I'm not against refactoring the `TabBar` if we can support dynamic tab width while keeping the current set of features.

Louies89:
@sospedra 
I am not sure, but, I guess, by using ViewPropTypes of RN, this can be solved. I saw it from[ here](https://github.com/godness84/react-native-recyclerview-list/blob/bc5c9489a574e8911a79422a0f919d044cd729fb/src/RecyclerViewList.js#L8)

pribeh:
I can help test as we're looking for this as well.

gabrielvcbessa:
I can help as well

jaitd:
I just created PR #544 which is our take on implementing dynamic width TabBar

satya164:
Hey, I just released a new alpha `2.0.0-alpha.0` of the library. It's rewritten using [react-native-gesture-handler](https://github.com/kmagiera/react-native-gesture-handler) and [react-native-reanimated](https://github.com/kmagiera/react-native-reanimated) addresses a many platform specific bugs and performance problems. The documentation is updated as well.

Please try the new version and see if it addresses your issue. If not, please open a new issue following the issue template.

PR #544, opened 2018-06-15
WIP: Dynamic width in TabBar

jaitd:
### Motivation

We use `react-native-tab-view` to show a swipeable menu for restaurants. The problem that we face is that the section names in the `TabBar` can be of varying length along with the fact that the `TabBar` needs to be scrollable. We found this RFC for Dynamic tab width in PR #356 but it hasn't seen updates for a while. We are already using the implementation from this pull request in our production app and haven't heard any complaints except the issue mentioned below.

### Implementation

In this pull request, we present our implementation to enable dynamic tab width in the `TabBar`. The feature can be enabled by setting the `dynamicWidth` prop for `TabBar`. 

`TabBar` listens to `onLayout` events for `View` components for a tab to find out the width of tab on the screen. These widths are stored in a map, which is in turn used to perform all the calculations for animations. We animate both the x-movement as well as the width of the indicator which leads to a nice visual effect.

### Remaining Issue

Currently, the `TabBar` scroll position is not kept in sync with the pan position received from the pager. We experimented with different pagers but in all cases the performance takes a large hit and the animation becomes very choppy. Our guess is that, the pan listener is fired a lot of times while panning on the pager which in turn leads to a lot of `scrollTo` calls on the `ScrollView`. @satya164, would you have any suggestions to improve this?

### Test plan

The example folder has been updated to include a page which showcases this feature. You can see it in action on Snack: [link](https://snack.expo.io/@jaitdixit/github.com-orda-technologies-react-native-tab-view:example@develop). Let me know if you want anything tested more specifically.

### DIFF
--- a/example/src/App.js
+++ b/example/src/App.js
@@ -19,6 +19,7 @@ import BottomBarIconTextExample from './BottomBarIconTextExample';
 import NoAnimationExample from './NoAnimationExample';
 import CoverflowExample from './CoverflowExample';
 import NativeDriverExample from './NativeDriverExample';
+import DynamicWidthExample from './DynamicWidthExample';
 
 const PERSISTENCE_KEY = 'index_persistence';
 
@@ -29,6 +30,7 @@ const EXAMPLE_COMPONENTS = [
   NoAnimationExample,
   CoverflowExample,
   NativeDriverExample,
+  DynamicWidthExample,
 ];
 
 type State = {
--- a/example/src/DynamicWidthExample.js
+++ b/example/src/DynamicWidthExample.js
@@ -1,0 +1,106 @@
+/* @flow */
+
+import * as React from 'react';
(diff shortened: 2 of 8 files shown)


erickreutz:
It's quite unfortunate that a PR like this doesn't even get a comment from maintainers. @jaitd is this the solution you ended up going with?

PierreCapo:
Any news about this PR @jaitd ? It would be a really useful addition to the lib.

andrictham:
Any updates on this?

jaitd:
Apologies for the late reply.

@erickreutz This is the solution that we went for in our app. Except for the issues mentioned in the pull request, we have not encountered any major problems.

@PierreCapo @andrictham I'm still waiting to hear from @satya164.

nehacurefit:
What is the plan for merging this PR? This is really great and useful

satya164:

PR #554, opened 2018-07-04
Selected tab bar indicator jumps to wrong positions  when tab bar is scrollable

wangbourne:
### Motivation
Everything is perfect if scene rendering is not expensive.  However, it comes a problem if that's expensive rendering, especially on old devices like iPhone5/5s and together with `ListView/FlatList`. I guess the root cause is expensive rendering could make `_renderIndicator transform` limp because of single js thread.

problem

This PR is to resolve a problem: selected tab bar indicator could goes to wrong positions.

### Test plan
Verified on old iPhone5 and new iPhoneX device


satya164:

PR #582, opened 2018-08-14
Fixed quick start to the snack example

jchorl:
### Motivation

The quick start code does not run. The styles variable is not defined, the tabs overlap the status bar, etc. The Snack code actually works and is pretty minimal, so we should just use that as the quick start.

### Test plan

Tested on my device. It runs.


satya164:

PR #587, opened 2018-08-20
Vertical Tab Bar (v2)

DaKaZ:
### Motivation

We have the desire/need to implement a vertical tab bar, primarily for use on Tablet and web (with react-native-web) layouts.

verticaltabbarexample

This PR implements two NEW components: `TabViewVertical` and `TabBarVertical` in hopes of overcoming the issues associated with the original attempt: PR #418 

### Test plan

Run the example app, there is a Vertical Tab Bar example there now.

### DIFF
--- a/example/src/App.js
+++ b/example/src/App.js
@@ -19,6 +19,7 @@ import BottomBarIconTextExample from './BottomBarIconTextExample';
 import NoAnimationExample from './NoAnimationExample';
 import CoverflowExample from './CoverflowExample';
 import NativeDriverExample from './NativeDriverExample';
+import VerticalTabBarExample from './VerticalTabBarExample';
 
 const PERSISTENCE_KEY = 'index_persistence';
 
@@ -29,6 +30,7 @@ const EXAMPLE_COMPONENTS = [
   NoAnimationExample,
   CoverflowExample,
   NativeDriverExample,
+  VerticalTabBarExample,
 ];
 
 type State = {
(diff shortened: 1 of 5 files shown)


satya164:
Thanks for the PR. As I mentioned in the previous PR, I would accept any required changes to make it easier to add this feature in userland, but I don't want to maintain the vertical tab bar.

Regarding the code in this PR, I did a diff of `TabViewVertical` with the current `TabView` to figure out the changes needed. The first change I noticed was that there was `sceneLayout` and `windowLayout` instead of just `layout`, however `windowLayout` is never used for anything. After removing it, the only change I see is this:

image

The style can be applied by passing a `style` prop to `TabView` without having to duplicate the code. So the changes seem unnecessary.

Regarding the tab bar, I don't want the vertical tab bar in the core, because:
- It's the most complex component in the library. Having to maintain 2 tab bars is too much for me. It'll be much better if you publish the tab bar as a different package and maintain it. The API contract is fairly stable, so updates to react-native-tab-view are unlikely to break it.
- The UI pattern feels alien to me, I have never seen such an UI pattern in official material design guidelines which this library aims to implement. I haven't seen this UI pattern in  any other design guideline or mobile apps either. I won't expect the pages to slide horizontally when I'm switching between them with a sidebar, because in a design sense they'll be located vertically.

Hence I won't be merging this PR. However, if there are any changes needed to the library to make it easier to implement this feature in userland, I'm happy to discuss it. For now, it seems like the only change required is `flexDirection: 'row'`, which you can pass in the `style` prop.

DaKaZ:
Per the discussion here, for anyone looking for vertical tab functionality I have published: https://github.com/DaKaZ/react-native-vertical-tab-view

Issue #468, opened 2018-02-28
TopTabBar with scrollEnabled causes indicator to render in wrong position on orientation


satya164:
https://github.com/react-navigation/react-navigation/issues/3534

DaKaZ:
I am working on tracking this one down.  Its easy to see the problem without changing orientations if you make the tab width's larger, like 320.  Its even more exaggerated when the initial index is 2.  As seen here, the indicator SHOULD be under the `ALBUMS` but its rendered to the far right.  

![screen shot 2018-08-20 at 2 03 27 pm](https://user-images.githubusercontent.com/895892/44363688-1f24a880-a482-11e8-8563-72a01c67c9dc.png)

In calculating the indicator transform, we have to take into account the amount scrolled already.  Trying to figure that part out now, animated values are a little foreign to me :)

satya164:
Hey, I just released a new alpha `2.0.0-alpha.0` of the library. It's rewritten using [react-native-gesture-handler](https://github.com/kmagiera/react-native-gesture-handler) and [react-native-reanimated](https://github.com/kmagiera/react-native-reanimated) addresses a many platform specific bugs and performance problems. The documentation is updated as well.

Please try the new version and see if it addresses your issue. If not, please open a new issue following the issue template.

PR #588, opened 2018-08-20
Fixes issue #468

DaKaZ:
### Motivation

This PR fixes the issues discussed in #468 as well as a secondary issue where the indicator will not render correctly if the tab width is large and the initial index requires a scroll.  For example, in the `TopBarTextExample.js` file if you change the state to `index: 2` and then the style to `tab: { width: 320 }` you can see this second problem.

The initially reported problem is resolved by a new `resetScroll` function which is called with width is changed.  The secondary issue is actually resolved by removing the `initialOffset` state - I am not entirely sure why this was messing that up nor am I am sure why the initial offset is needed at all since all the examples still work as desired without it.

### Test plan

Run the examples, change orientation, see the indicator stick with the selected tab.

NOTE: the tab layout can still be slightly off after an orientation change, but at least now the indicator and the tabs are offset together.

### DIFF
--- a/src/TabBar.js
+++ b/src/TabBar.js
@@ -46,7 +46,6 @@ type Props<T> = SceneRendererProps<T> & {
 type State = {|
   visibility: Animated.Value,
   scrollAmount: Animated.Value,
-  initialOffset: ?{| x: number, y: number |},
 |};
 
 const useNativeDriver = Boolean(NativeModules.NativeAnimatedModule);
@@ -91,21 +90,9 @@ export default class TabBar<T: *> extends React.Component<Props<T>, State> {
       }
     }
 
-    const initialOffset =
-      this.props.scrollEnabled && this.props.layout.width
-        ? {
-            x: this._getScrollAmount(
-              this.props,
-              this.props.navigationState.index
-            ),
-            y: 0,
-          }
-        : undefined;
-
     this.state = {
       visibility: new Animated.Value(initialVisibility),
       scrollAmount: new Animated.Value(0),
-      initialOffset,
     };
   }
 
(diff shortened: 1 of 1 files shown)


src/TabBar.js
```diff
@@ -153,6 +140,18 @@ export default class TabBar<T: *> extends React.Component<Props<T>, State> {
   _panXListener: string;
   _offsetXListener: string;
 
+  _resetState = () => {
+    this.props.scrollEnabled && this._stopTrackingPosition();
+    this.setState(
```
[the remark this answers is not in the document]
DaKaZ: Yes, that does work, great suggestion.

DaKaZ:
@satya164 please see the latest changes that address your review.  Thanks

satya164:
I'll try to check this weekend. Thanks a lot :D

src/TabBar.js
```diff
@@ -155,6 +142,12 @@ export default class TabBar<T: *> extends React.Component<Props<T>, State> {
   _panXListener: string;
   _offsetXListener: string;
 
+  _resetState = () => {
+    this.props.scrollEnabled && this._stopTrackingPosition();
```
satya164: Don't think stopping and starting tracking position again is necessary here

src/TabBar.js
```diff
@@ -155,6 +142,12 @@ export default class TabBar<T: *> extends React.Component<Props<T>, State> {
...
+    this.state.scrollAmount.setValue(0);
```
satya164: How does this fix the issue? If the current tab is not the first one, shouldn't this value be set accordingly?

satya164:
Which platform is it?

z252383898:
version 1.3.1 has new problem, when you first open and it need scroll , the indicator well scroll the wrong position

satya164:

Issue #586, opened 2018-08-19
renderLabel is missing `focused` and `index` parameters


t3chnoboy:
### Current behaviour

After this commit https://github.com/react-native-community/react-native-tab-view/commit/8116fc9963479630385f79ba27d054cef8e10dfe 
`renderLabel`, `renderIcon`, `renderBadge` methods no longer receive `focused` and `index` parameters.
### Expected behaviour

Multiple issues and PRs suggest that changing the tab elements based on the active tab is a very common problem and all of the solutions rely on the `index`/ `focused` props.
The solutions weren't very simple already but I understand that it's because of the animations.
After removing the `index` and `focused` parameters from the `renderLabel` method it seemed to become even harder and requires even more code to achieve the expected behavior.

Is there a way of changing the active tab label color without overwriting the whole TabBar render method like in one of your examples 
https://snack.expo.io/@satya164/custom-tabbar-example ?

Thanks!

DaKaZ:
In an effort to help with this issue, I have submitted a small PR that will pass the `focused` prop to all three methods.  See https://github.com/react-native-community/react-native-tab-view/pull/589

flosdor:
I used the renderLabel method.

`
renderLabel(props) {
  let index = 0;
  return ({ route }) => {
    const focused = index === props.navigationState.index;
    index += 1;
    return (
      <View>
        <Text
          style={[
            styles.labelStyle,
            focused ? styles.labelSelectedStyle : null,
          ]}
        >
          {route.title}
        </Text>
      </View>
    );
  };
}
`
and
`
<TabBar
    ...
    renderLabel={this.renderLabel(props)}
/>
`

satya164:
Hey, I just released a new alpha `2.0.0-alpha.0` of the library. It's rewritten using [react-native-gesture-handler](https://github.com/kmagiera/react-native-gesture-handler) and [react-native-reanimated](https://github.com/kmagiera/react-native-reanimated) addresses a many platform specific bugs and performance problems. The documentation is updated as well.

Please try the new version and see if it addresses your issue. If not, please open a new issue following the issue template.

PR #589, opened 2018-08-20
fixes issue #586 by passing `focused` to various render methods

DaKaZ:
### Motivation

This is a small PR that adds the `focused` prop to the `renderLabel`, `renderIcon` and `renderBadge` function calls so that upstream functions can easily know that what they want is currently focused.

### Test plan

Use the code, see the `focused` prop passed to you

### DIFF
--- a/src/TabBar.js
+++ b/src/TabBar.js
@@ -13,7 +13,7 @@ import {
 } from 'react-native';
 import TouchableItem from './TouchableItem';
 import { SceneRendererPropType } from './PropTypes';
-import type { Scene, SceneRendererProps } from './TypeDefinitions';
+import type { Scene, SceneRendererProps, RenderProps } from './TypeDefinitions';
 import type {
   ViewStyleProp,
   TextStyleProp,
@@ -32,9 +32,9 @@ type Props<T> = SceneRendererProps<T> & {
   getAccessible: (scene: Scene<T>) => ?boolean,
   getAccessibilityLabel: (scene: Scene<T>) => ?string,
   getTestID: (scene: Scene<T>) => ?string,
-  renderLabel?: (scene: Scene<T>) => React.Node,
-  renderIcon?: (scene: Scene<T>) => React.Node,
-  renderBadge?: (scene: Scene<T>) => React.Node,
+  renderLabel?: (scene: RenderProps<T>) => React.Node,
+  renderIcon?: (scene: RenderProps<T>) => React.Node,
+  renderBadge?: (scene: RenderProps<T>) => React.Node,
   renderIndicator?: (props: IndicatorProps<T>) => React.Node,
   onTabPress?: (scene: Scene<T>) => mixed,
   tabStyle?: ViewStyleProp,
@@ -188,7 +188,7 @@ export default class TabBar<T: *> extends React.Component<Props<T>, State> {
     this._adjustScroll(value);
   };
 
-  _renderLabel = (scene: Scene<*>) => {
+  _renderLabel = (scene: RenderProps<*>) => {
     if (typeof this.props.renderLabel !== 'undefined') {
       return this.props.renderLabel(scene);
     }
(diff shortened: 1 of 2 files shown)


satya164:
Thanks for the PR, the reason there is no such argument passed to these functions is because it doesn't work with animations. The recommended way is to use the `position` prop, it's a bit more work, but it ensures the tab transitions are animated. And if you really need it, since you own the state, it's just a matter of doing `const focused = this.state.routes.indexOf(route) === this.state.index`.

But I agree we can make styling focused and unfocused state easier without having to deal with `position` prop. The way to do t would be to absolutely position the focused and unfocused labels/icons on top of each other and crossfade them, and then we can pass a `focused` prop.

DaKaZ:
@satya164 I think I follow, but before I do an update to the PR, lets chat and make sure.  

What I hear you saying is that we should allow a for user to pass in both a focused and unfocused style like: `this.props.labelStyle` and `this.props.labelStyleFocused`.  Assuming we have both of those, we can then render both the focused and unfocused label, icon and badge, then use position to cross fade the two as a tab becomes focused.  

The problem I see is that `this.props.renderIcon({ route })` and `this.props.renderBadge({ route })` don't have local render options like label and we don't pass the current state to these functions (ie, index or focused).  

So we need to either pass the `opacity` animated value const (see line 402) to those methods OR we need to pass the full `navigationState` or `navigationState.index` or `focused` to the user space render functions in order to render the icons and badges in the focused vs unfocused state.

Does that sound right?

satya164:
> we should allow a for user to pass in both a focused and unfocused style like: this.props.labelStyle and this.props.labelStyleFocused. Assuming we have both of those, we can then render both the focused and unfocused label, icon and badge, then use position to cross fade the two as a tab becomes focused

I'd use `activeLabelStyle` and `inactiveLabelStyle`, but yes

> The problem I see is that this.props.renderIcon({ route }) and this.props.renderBadge({ route }) don't have local render options like label and we don't pass the current state to these functions (ie, index or focused).

When we start cross-fading them, we can pass a `focused` prop. Then we'll call each function for both unfocused and focused states with `focused` set to `true` and `false`. I think it'll be easier to use if we manage the opacity ourselves.

santiagofm:
Hey guys! any update on this PR?

luissmg:
@DaKaZ I think both solutions have an issue. _You just have to choose what fits the better in this case.
If you keep the current prop, then you will have kind of a redundancy in that prop and you will need to do some verifications.
If you replace the current prop, old code that people are using could, indeed, break. But hey, if it is a new release, it is expect that some things can change. If people want the other version they can do it anyway specifying the version on `package.json`.

DaKaZ:
I have not worked on this yet mainly because most of my PRs to this code are rejected by @satya164 - I am happy to do the work but would like to know from him if the approach is acceptable before I spend the time.

IvanLinos:
> I'd use `activeLabelStyle` and `inactiveLabelStyle`, but yes

Yeah! That would be an ideal solution if inactiveLabelStyle was an actual prop... ;)

Could you please confirm @DaKaZ if this solution, as a PR, will be implemented if done? (by him, me or whoever will code it).

satya164:
> We currently have the prop labelStyle - switching to your suggestion will break everyone's implementation or worse, if we leave labelStyle and add the two new ones, the props become quite ambiguous. I think it would be better to preserve the existing API and just add activeLabelStyle as an optional prop and if its included, then we do the crossfade.

IMO `labelStyle` should be always applied to the label and `activeLabelStyle` and `inactiveLabelStyle` should override it in active and inactive states. Then you can do something like set `opacity` to `1` with `labelStyle` and then change the color using `activeLabelStyle` or `inactiveLabelStyle`. If you think it's ambiguous, documenting this behaviour should avoid any confusion.

luissmg:
@satya164 I think that's a good solution. Are you going to continue de development or not @DaKaZ?

MickaelNeves:
Any update on this PR??

satya164:

vohoangankhanh:
any update on this, ??

ereenyessamhanna:
any updates ?

MinaFSedrak:
any update ?

PR #602, opened 2018-09-04
Add support for RTL in TabView

horodnicdragos:
### Motivation

This pull requests adds support for RTL problems described in the following issue: #184

Animations, indicator and focus are off when in RTL. Notice the indicator doesn't change when moving between tabs and label focus is off as well.

The culprit for most issues seems to be in calculating the position in TabView.

### Test plan

- Test on RTL configured device (e.g. Arabic)
- Use Android's Developer Options -> Force RTL
- Add code in the app: I18nManager.forceRTL(true)

### DIFF
--- a/src/PagerAndroid.js
+++ b/src/PagerAndroid.js
@@ -120,15 +120,11 @@ export default class PagerAndroid<T: *> extends React.Component<Props<T>> {
 
   render() {
     const { navigationState, swipeEnabled, keyboardDismissMode } = this.props;
-    let children = [];
-    React.Children.forEach(this.props.children, c => {
-      if (I18nManager.isRTL) {
-        children = [c, ...children];
-      } else {
-        children = [...children, c];
-      }
-    });
-    const content = React.Children.map(children, (child, i) => {
+    const children = I18nManager.isRTL
+    ? React.Children.toArray(this.props.children).reverse()
+    : React.Children.toArray(this.props.children);
+    
+    const content = children.map((child, i) => {
       const route = navigationState.routes[i];
       const focused = i === navigationState.index;
 
--- a/src/PagerAndroid.js
+++ b/src/PagerAndroid.js
@@ -33,8 +33,7 @@ export default class PagerAndroid<T: *> extends React.Component<Props<T>> {
 
   componentDidUpdate(prevProps: Props<T>) {
     if (
-      prevProps.navigationState.routes.length !==
-        this.props.navigationState.routes.length ||
+      prevProps.navigationState.routes !== this.props.navigationState.routes ||
       prevProps.layout.width !== this.props.layout.width
     ) {
       this._handlePageChange(this.props.navigationState.index, false);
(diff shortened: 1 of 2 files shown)


src/PagerAndroid.js
```diff
@@ -122,12 +119,15 @@ export default class PagerAndroid<T: *> extends React.Component<Props<T>> {
...
   render() {
-    const {
-      children,
-      navigationState,
-      swipeEnabled,
-      keyboardDismissMode,
-    } = this.props;
+    const { navigationState, swipeEnabled, keyboardDismissMode } = this.props;
+    let children = [];
+    React.Children.forEach(this.props.children, c => {
```
satya164: Isn't this `.forEach` function the same as `.reverse`, which we were doing previously with `content.reverse()`?

satya164: Yeah, but doesn't that mean we just need to reverse before mapping and not after? Because what you're doing with `forEach` is just `reverse`:

```js
var items = ['a', 'b', 'c'];
var result = [];

items.forEach(c => result = [c, ...result]);

console.log(result); // ["c", "b", "a"]

items.reverse();

console.log(items); // ["c", "b", "a"]
```

horodnicdragos:
Hey there @satya164! Does it look good to you for a merge?

src/PagerAndroid.js
```diff
@@ -33,8 +33,7 @@ export default class PagerAndroid<T: *> extends React.Component<Props<T>> {
 
   componentDidUpdate(prevProps: Props<T>) {
     if (
-      prevProps.navigationState.routes.length !==
-        this.props.navigationState.routes.length ||
+      prevProps.navigationState.routes !== this.props.navigationState.routes ||
```
satya164: Hey, why did we need this change?

satya164:
@horodnicdragos thanks! I'll merge and release this weekend. left one question

src/PagerAndroid.js
horodnicdragos: Does the length change over time? Aren't the references different when the routes change?

satya164: @horodnicdragos the length could change if the user is adding/removing tags dynamically. the routes array could change even if the number of routes didn't change, but say the user added a new property/changed some value in a specific route. we don't need to update the scroll position in that case.

satya164:
merged in 815557e1f36b05b284b982ba0a12a9e5b88431d1

src/PagerAndroid.js
horodnicdragos: I understand now, it makes sense. Let me revert those changes.

horodnicdragos:
Did you revert the unnecessary changes you've mentioned above already?

src/PagerAndroid.js
satya164: No need. Already did it and merged.

noambonnie:
Thank you both for this work. @horodnicdragos sorry I wasn't able to pick this up. I was busy here (aren't we all though!).

592da:
any updates ?

PR #603, opened 2018-09-07
Update tabbar to be more performant if you are using ios and nativeDriver

derekbar90:
### Motivation

Currently am using the experimental pager and have found that the tabbar is lagging when scrollEnabled is set to true (ios only) when scrolling through tabs. To fix this i saw that we were setting animation on the scrollTo which was causing this issue. 

### Test plan

Create 4+ tabs which is 100% the screen width and have a long enough title so that the list of items will cause a scroll on the TabBar. Navigate to a new tab. View the lag between when the indicator is and what is present on the screen

### DIFF
--- a/src/PagerExperimental.js
+++ b/src/PagerExperimental.js
@@ -9,6 +9,7 @@ import type { PagerRendererProps } from './TypeDefinitions';
 type Props<T> = PagerRendererProps<T> & {
   swipeDistanceThreshold?: number,
   swipeVelocityThreshold?: number,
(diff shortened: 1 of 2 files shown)


nenti:
I can confirm heavy lag on the experimental with scrolltabs on ios.

derekbar90:
@nenti I'll update this MR with proper linking and maybe we can get a review. Have you tried pulling this branch to test the difference? I'm using it in production now.

satya164:

PR #631, opened 2018-10-27
better performance  (Delay rendering of non-focused tabs to next tick)

Niryo:
### Motivation

Delay rendering of non focused tabs to the next tick, for better performance.
In order to keep it opt-in and fully backward compatible, I decided to add a new prop: `delayRenderOfNonFocusedTabs` to `TabView`.
If you set this prop to true, your current tab will be rendered, and only on the next tick the other tabs will be rendered.
edge case: If you turn this option on, and the render time of the current tab is long, and the user is expected to switch to other tab very quickly, then he will see a blank tab for a split second. that's why this option is currently opt-in.

### Results:
When clicking on a group called "blamos2", you can see how much time it takes to load the group items in a page with many tabs, before and after the change 

|Before|After|
|---|---|
| ezgif com-resize|ezgif com-resize 1   |

### DIFF
--- a/src/TabView.js
+++ b/src/TabView.js
@@ -26,6 +26,7 @@ type Props<T> = PagerCommonProps<T> &
     renderTabBar: (props: SceneRendererProps<T>) => React.Node,
     tabBarPosition: 'top' | 'bottom',
     useNativeDriver?: boolean,
+    delayRenderOfNonFocusedTabs?: boolean,
     style?: ViewStyleProp,
   };
 
@@ -35,6 +36,7 @@ type State = {|
   panX: Animated.Value,
   offsetX: Animated.Value,
   position: any,
+  delayRenderOfNonFocusedTabs: any,
 |};
 
 export default class TabView<T: *> extends React.Component<Props<T>, State> {
@@ -49,6 +51,7 @@ export default class TabView<T: *> extends React.Component<Props<T>, State> {
     renderPager: PropTypes.func.isRequired,
     renderScene: PropTypes.func.isRequired,
     renderTabBar: PropTypes.func,
+    delayRenderOfNonFocusedTabs: PropTypes.bool,
     tabBarPosition: PropTypes.oneOf(['top', 'bottom']),
   };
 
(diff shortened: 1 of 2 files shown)


src/TabView.js
```diff
@@ -189,15 +189,18 @@ export default class TabView<T: *> extends React.Component<Props<T>, State> {
...
                 ...props,
                 route,
               });
 
               if (React.isValidElement(scene)) {
                 /* $FlowFixMe: https://github.com/facebook/flow/issues/4775 */
-                return React.cloneElement(scene, { key: route.key });
+                return React.cloneElement(scene, {
+                  key: route.key,
+                  isFocused: navigationState.index === index,
```
satya164: Please remove this. I don't want to add this prop. It'll cause re-render every tab change and defeats the purpose of using `SceneMap` or `PureComponent`

src/SceneMap.js
```diff
@@ -2,10 +2,24 @@
...
+type State = {|
+  shouldRenderContent: boolean,
+|};
+
+class SceneComponent extends React.PureComponent<*, State> {
+  constructor(props) {
+    super(props);
+    this.state = { shouldRenderContent: false };
+    //delay rendering of not focused tabs to the next tick:
+    setTimeout(() => this.setState({ shouldRenderContent: true }), 0);
```
satya164: Can you move this code to `TabView`? This can be done for all scenes, not only the ones using `SceneMap`

src/SceneMap.js
```diff
@@ -2,10 +2,24 @@
...
+  }
   render() {
     const { component, ...rest } = this.props;
-    return React.createElement(component, rest);
+    if (this.state.shouldRenderContent || rest.isFocused) {
```
satya164: This will cause a visible empty screen. Have you tried if it works properly when animations and gestures are enabled?

Niryo: The visible empty screen is only for one tick, on not focused tab, so it is not actually visible, unless if I am missing something. When you say animations and gestures, do you mean something different from what we are seeing in the attached gif?

satya164: you're right. for some reason I thought it was `&&`, not `||`

src/SceneMap.js
satya164: @Niryo you can add a variable to state indicating if it's initial render, and then add a check like `this.state.initial || i === navigationState.index` here: https://github.com/react-native-community/react-native-tab-view/blob/master/src/TabView.js#L193

Niryo:
@satya164, as you suggested, I added this change to the tabView. I also edited the description of this PR, please read it:)

satya164:
I'll check soon. Thank you!

Niryo:
@satya164 I added the props to the docs also.

satya164:
Thank you! Merged in 5b0a661b24ea587a0e4c282c6cbbd553d402877e

I have removed the prop for now and enabled it by default. I'll add the prop if people want it.

PR #640, opened 2018-11-16
Let TAB switch execute early.

fantasy525:
### Motivation

I found that TAB switching didn't start until the animation of the page had stopped completely, which made it feel like TAB switching was slow. I wanted to start TAB switching when my hand left the screen and the page had moved to the next page.
here is slow:
slow

here is fast:
fast


satya164:

PR #675, opened 2019-02-03
Fix RTL tabview indicator position

MahmoudAliIbrahim:
This small change fixes indicator position for rtl layout with scroll-enabled on tabbar
indicator is not correctly aligned with the tab:
issue:
ios-old
fix:
ios-correct

code fixes the issue for ios, but for android i'm still having issues, maybe someone can help.


MahmoudAliIbrahim:
closed until fixing the issue for android

PR #681, opened 2019-02-16
BREAKING CHANGE: rewrite using gesture handler and reanimated

satya164:
**This is a work in progress**

Demo: https://exp.host/@satya164/react-native-tab-view-demos

## TODO

Below is the list of tasks I need to finish to release a beta:

- [ ] Polish the spring animation during swipe
- [x] Use a faster animation when switching tabs on tab press
- [x] Make sure the `position` prop is usable on first render until layout is measured
- [ ] Figure out how to make the pager play well with scrollview
- [x] Fix an issue on iOS where something is `NaN` after tab press
- [ ] Fix bugs (???)

## Motivation

The PR aims to implement the tab view using gesture handler and reanimated to solve long-standing performance issues and platform inconsistencies. It's been very annoying to deal with platform differences in `ScrollView` and buggy behavior in `ViewPagerAndroid`.

The rewrite uses a single implementation for both iOS and Android. This should be much more maintainable and will have consistent behavior across platforms.

I have tried my best to keep the API mostly unchanged for an easier upgrade, but breaking changes are inevitable. The good news is, if you don't use a custom tab bar or pager, the migration should require minimal changes (other than linking gesture handler and reanimated)

## Breaking changes

- The `renderPager` prop is removed. Now there's only one pager implemented with gesture handler.
- The `animationEnabled` prop is removed. It's not straightforward to implement, and doesn't make much sense in the scope of this library.
- The `canJumpToTab` prop is removed
- Few props such as `onSwipeStart`, `onSwipeEnd` and `onAnimationEnd` have been removed (if you have ideas on how to implement these, happy to add them back)
- The `position` prop received by the `renderTabBar` and `renderScene` are now [reanimated](https://github.com/kmagiera/react-native-reanimated) nodes, so you'll need to migrate your custom tab bar to use reanimated. This should mostly be straightforward.
- Any scrollable content inside the tabs needs to use `ScrollView`/`FlatList` etc. from [gesture handler](https://github.com/kmagiera/react-native-gesture-handler)
- No react-native-web support because both gesture handler and reanimated are native deps. I can implement the pager with pan responder for web, but can't do much for reanimated

tab-view-small


example/src/CoverflowExample.js
```diff
@@ -71,17 +59,17 @@ export default class CoverflowExample extends React.Component<*, State> {
       }
     });
 
-    const translateX = position.interpolate({
+    const translateX = Animated.interpolate(position, {
```
osdnk: and related. 

You may consider using `position.interpolate` with `reanimated` as well

src/Pager.js
```diff
@@ -0,0 +1,282 @@
...
+  _isSwipeGesture = new Value(FALSE);
+
+  _clock = new Clock();
+
+  _routesLength = new Value(this.props.navigationState.routes.length);
+
+  // tslint:disable-next-line: strict-boolean-expressions
+  _layoutWidth = new Value(this.props.layout.width || 320);
+
+  _swipeDistanceThreshold = divide(this._layoutWidth, 1.75);
```
osdnk: maybe it could be configurable somehow?

src/Pager.js
```diff
@@ -0,0 +1,282 @@
...
+          // We weren't dragging before, set it to true
+          set(this._isSwiping, TRUE),
+          set(this._isSwipeGesture, TRUE),
+          // Also update the drag offset to the last position
+          set(this._offsetX, this._position),
+        ]),
+        // Update position with previous offset + gesture distance
+        set(this._position, add(this._offsetX, this._gestureX)),
+        // Stop animations while we're dragging
+        stopClock(this._clock),
```
osdnk: I think it could be `cond(clockRunning(this._clock), stopClock(this._clock))` I think it will be clearer then what's the purpose of this line.

src/Pager.js
```diff
@@ -0,0 +1,282 @@
...
+                    this._index,
+                    cond(
+                      greaterThan(
+                        // Gesture can be positive, or negative
+                        // Get absolute for comparision
+                        abs(this._gestureX),
+                        this._swipeDistanceThreshold
+                      ),
+                      // If gesture value exceeded the threshold, calculate direction from distance
+                      divide(this._gestureX, abs(this._gestureX)),
```
osdnk: What's the reason for making it so hacky? Dividing on vary small integers in Java may give some unwanted effects (heard on Uni xd) and it's a bit more expensive. Of course, it probably does not matter, but if I were you I would use `cond(greatedThan(_gestureX, 0), 1, -1)`

satya164: I copied it from my previous impl where it is JS. Will change it :)

example/src/CoverflowExample.js
satya164: Does the method exist? It didn't in the version I'm using at least (expo)

src/Pager.js
satya164: Yeah, will make it configurable

satya164:
I've renamed master to 1.0 and merged this branch to master. I also released `2.0.0-alpha.0` to npm, so you can try it in your apps.

PR #689, opened 2019-02-24
feat: Adding props to override minHeight

Tonioverzeworld:
Due to some needs on a project I needed to be able to customize the minHeight of the tabBar but for now it was sealed so I decided to add tabBarItemStyle prop in order to be able to customize the item

### Motivation

The TabBar minHeight was not customizable and added always a 10 padding. It seems not possible to resolve this problem with a different approach as there is no style customization option for this element. 

### Test plan

on ScrollableTabBarExample.js on line 48 add the following
```js
tabBarItemStyle={{ minHeight: 0, padding: 0 }}
````
image-1

### DIFF
--- a/src/TabBarItem.js
+++ b/src/TabBarItem.js
@@ -38,6 +38,7 @@ type Props<T> = {|
   onLongPress: () => mixed,
   tabWidth: number,
   labelStyle?: TextStyleProp,
+  tabBarItemStyle: ViewStyleProp,
   style: ViewStyleProp,
 |};
 
@@ -61,6 +62,7 @@ export default function TabBarItem<T: Route>({
   pressColor,
   pressOpacity,
   labelStyle,
+  tabBarItemStyle,
   style,
   tabWidth,
   onPress,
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: src/TabBarItem.js [hunk 1]
```diff
@@ -201,7 +203,10 @@ export default function TabBarItem<T: Route>({
       onLongPress={onLongPress}
       style={tabContainerStyle}
     >
-      <View pointerEvents="none" style={[styles.item, itemStyle]}>
+      <View
+        pointerEvents="none"
+        style={[styles.item, itemStyle, tabBarItemStyle]}
```
satya164: I forgot to pass `tabStyle` here. Can you change this to pass it instead? Then you won't need the new prop.

Tonioverzeworld: oh great, I'll make this change and submit new pull request today ! thanks

satya164: no worries, I just made the change and did a release

Tonioverzeworld: oh great thanks

PR #701, opened 2019-03-04
Add activeTabStyle

svbutko:
### Motivation

There's no proper way to change the container style of a focused item.

### Test plan

1. Add `activeTabStyle` with a distinguish style to `TabBar`
2. Select a tab

### DIFF
--- a/src/TabBarItem.js
+++ b/src/TabBarItem.js
@@ -38,6 +38,7 @@ type Props<T> = {|
   onLongPress: () => mixed,
   tabWidth: number,
   labelStyle?: TextStyleProp,
+  activeTabContainerStyle?: ViewStyleProp,
   style: ViewStyleProp,
 |};
 
@@ -61,6 +62,7 @@ export default function TabBarItem<T: Route>({
   pressColor,
   pressOpacity,
   labelStyle,
+  activeTabContainerStyle = {},
   style,
   tabWidth,
   onPress,
@@ -164,13 +166,13 @@ export default function TabBarItem<T: Route>({
     (tabStyle && typeof tabStyle.width !== 'undefined') ||
     scrollEnabled === true;
 
-  const tabContainerStyle = {};
+  let tabContainerStyle = isFocused ? activeTabContainerStyle : {};
   const itemStyle = isWidthSet ? { width: tabWidth } : null;
 
   if (tabStyle && typeof tabStyle.flex === 'number') {
-    tabContainerStyle.flex = tabStyle.flex;
+    tabContainerStyle = StyleSheet.flatten([{flex: tabStyle.flex}, tabContainerStyle]);
   } else if (!isWidthSet) {
-    tabContainerStyle.flex = 1;
+    tabContainerStyle = StyleSheet.flatten([{flex: 1}, tabContainerStyle]);
   }
 
   const scene = { route };
--- a/src/TabBarItem.js
+++ b/src/TabBarItem.js
@@ -38,7 +38,7 @@ type Props<T> = {|
   onLongPress: () => mixed,
   tabWidth: number,
   labelStyle?: TextStyleProp,
-  activeTabContainerStyle?: ViewStyleProp,
+  activeTabStyle?: ViewStyleProp,
   style: ViewStyleProp,
 |};
 
@@ -62,7 +62,7 @@ export default function TabBarItem<T: Route>({
   pressColor,
   pressOpacity,
   labelStyle,
-  activeTabContainerStyle = {},
+  activeTabStyle = {},
   style,
   tabWidth,
   onPress,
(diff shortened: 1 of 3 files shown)


longunibbo:
I'm very looking forward to seeing this. How can I use your implementation in my project without waiting for the update from the author of this library? Thank you in advance

svbutko:
@longunibbo replace the dependency from the package.json with this one:
    `"react-native-tab-view": "git+https://github.com/svbutko/react-native-tab-view.git#active-tab-container-style",`

satya164:
Thanks for the PR. Can you mention your use case? Why do you need to customize the active tab container style?

unigeniee:
@satya164 Can you implement something like this without customized style?
From the manual, I do not see any formal guide to do it
image

longunibbo:
@satya164 Can you implement something like this?
I do not see any formal guide to do it from the manual.
image

longunibbo:
@svbutko I'm trying to use your lib but I have a concerned about your lib behavior. It looks strange

image

This is my code

_renderTabBar1 = props => {

        return (
            <TabBar
                {...props}
                style={styles.tabbar}
                tabStyle={styles.tab}
                activeTabContainerStyle={styles.tabFocused}
                labelStyle={styles.label}
                elevation={0}
            />)
    };

const styles = StyleSheet.create({
    tab: {
        height: 45,
        flex: 1,
        alignItems: 'center',
        justifyContent: 'center',
        borderColor: '#511A52',
        borderWidth: 5,
        borderRadius: 45 / 2,
    },
    tabFocused: {
        flex: 1,
        height: 60,
        alignItems: "center",
        padding: 16,
        justifyContent: 'center',
        borderColor: 'green',
        borderWidth: 1,
        borderRadius: 30,
    }
});

svbutko:
> Thanks for the PR. Can you mention your use case? Why do you need to customize the active tab container style?

@satya164 There are cases several cases but primarily it's when one need to have a different background or border of a focused tab.
tabexample

svbutko:
@longunibbo can you try it now, I replaced **activeTabContaienrStyle** with **activeTabStyle**, it was my mistake

satya164:
I think if we add `activeTabStyle`, there should also be `inactiveTabStyle` (like we have `activeColor` and `inactiveColor`. Can you add the prop?

Frekansapp:
@satya164 could you merge this PR? I realy need this feature.

satya164:
@Frekansapp I have requested to add another prop and the CI is failing. I cannot merge before they are done.

bakaa99:
> I think if we add `activeTabStyle`, there should also be `inactiveTabStyle` (like we have `activeColor` and `inactiveColor`. Can you add the prop?

Are you high?? what will the use of inactiveTabStyle, when we have the default tabSyle prop, we just need a prop to customize the current selected tab.

satya164:
hey @bakaa99, maybe next time try to be nice to open source maintainers who are working for free on libraries that you are using.

svbutko:
Hi @satya164 I', sorry that I didn't reply for so long.
Basically `inactiveTabStyle` would be redundant and cause transition and ambiguity issues due to the fact that this repo already has `tabStyle` property which already does the same as `inactiveTabStyle`.

satya164:
> this repo already has tabStyle property which already does the same as inactiveTabStyle.

`tabStyle` applies to both active and inactive styles though. So if there's some style you only want to apply for inactive tabs, `activeTabStyle` needs to override those styles as well

> Basically inactiveTabStyle would be redundant

I think it's better to explicitly specify what styles you want for active and inactive styles. It also seems more obvious to have a `inactiveStyle` prop when there's a `activeStyle` prop looking at other APIs in the library. I'd like the API of the lib to be consistent.

> cause transition and ambiguity issues

Disagree with this, it's not possible to cause any transition issues with `inactiveTabStyle` that you couldn't with `tabStyle`

vikramkh:
How do I make the active label different from the inactive label if i use a custom render label?

FelipeGatti:
Hello!
Do you guys know when its going to be released ? 

Thanks in advance.

src/TabBarItem.js
```diff
@@ -167,8 +169,10 @@ export default function TabBarItem<T: Route>({
   const tabContainerStyle = {};
   const itemStyle = isWidthSet ? { width: tabWidth } : null;
 
+  const isActiveStyle = isFocused && activeTabStyle != null;
+
   if (tabStyle && typeof tabStyle.flex === 'number') {
-    tabContainerStyle.flex = tabStyle.flex;
+    tabContainerStyle.flex = isActiveStyle ? activeTabStyle.flex : tabStyle.flex;
```
jmarkstevens: Shouldn't this be isActiveStyle && typeof activeTabStyle.flex === 'number'...

Gethyl:
Hey @satya164 , 

Do you know if this issue will be revisited and the PR merged? Would really appreciate to have this added in. 

Right now the issue I have is, if I want to use my own `renderTabBar`, then it's easy to custom, but then I don't get the scroll. And the effort to get the scroll for my custom `renderTabBar` seems too much. But at the same time, if I use the `TabBar` scroll is really nice, but then I miss out the option to customise the selected tab :-(

Gethyl:
I managed to customise the tab by setting the padding and margin to 0 in the `tabStyle`, and for `renderLable` I return a view as shown below 

```
 <TabBar
        {...props}
        renderLabel={({ route, focused, color }) => (
          <View
            style={{
              backgroundColor: focused ? 'green' : 'transparent',
              height: 55,
              justifyContent: 'center',
              alignItems: 'center',
              paddingHorizontal: 20,
            }}>
            <Text style={{ color: focused ? 'white' : 'black', margin: 8 }}>
              {route.title}
            </Text>
          </View>
        )}
        scrollEnabled
        style={styles.tabBar}
        tabStyle={styles.tab}
        labelStyle={styles.label}
      />

![image](https://user-images.githubusercontent.com/15939823/71502796-408f7300-28c6-11ea-80e2-6fe7fa4a2c37.png)

anastely:
Hey, @Gethyl Thank you for this example,
I add some edits on it, but I have some issue in tab width, I don't know how can I get it, So I want to add the label background to fill the gap, 
I managed it by using Dimensions but I guess it does not work as expected in others cases

Issue #716, opened 2019-03-15
Why set index after animation is done?


lindesvard:
### Current behaviour

This is not a bug but the experience is poor with the current behaviour.
After the animation is done the new index is updated.

Similar issue: https://github.com/react-native-community/react-native-tab-view/issues/463

### Expected behaviour

When we actually know what the next index is, it should be updated.

### Code sample

https://snack.expo.io/Syl8kbtvV

### What have you tried

Haven't looked in the Pager component so much but why not do the `call([this._index], ([value]))` in the Pager component when the animation starts instead of finished? 

I tried to do the code below and works well. I needed to `this._currentIndexValue = value` as well otherwise it would jump 2 screens. I believe that you have a reason for doing it after the animation is done but I would like to know why. I can do a PR if I actually know the reason behind it.

```
  cond(clockRunning(this._clock), NOOP, [
        // Animation wasn't running before
        // Set the initial values and start the clock
        set(toValue, multiply(index, this._layoutWidth, DIRECTION_RIGHT)),
        set(frameTime, 0),
        set(state.time, 0),
        set(state.finished, FALSE),
        set(this._index, index),
        startClock(this._clock),
        call([this._index], ([value]) => {
          // If the index changed, and previous animation has finished, update state
          this._currentIndexValue = value
          this.props.onIndexChange(value)

          // Without this check, the pager can go to an infinite update <-> animate loop for sync updates
          if (value !== this.props.navigationState.index) {
            this._pendingIndexValue = value

            // Force componentDidUpdate to fire, whether user does a setState or not
            // This allows us to detect when the user drops the update and revert back
            // It's necessary to make sure that the state stays in sync
            this.forceUpdate()
          }
        }),
      ]),
```
### Your Environment

| software                     | version
| ---------------------------- | -------
| ios or android               | both
| react-native                 | 0.58.3
| react-native-tab-view        | 2.0.2
| react-native-gesture-handler | 1.1.0
| react-native-reanimated      | 1.0.0-alpha.12
| node                         | 10
| npm or yarn                  | yarn@1.6.0

putuyoga:
I got this experience too, especially after upgraded the lib to `v2.0.2`. Because `v1.3.2` is buggy on `react-native v0.59` (_the tab button press is not working anymore_). Any idea @satya164 ?

pontusab:
Would also like to have the index updated as soon as possible or at least before the animation is done.

satya164:
[...]

PR #718, opened 2019-03-18
fix: change index when we know the next index

lindesvard:
### Motivation

Regarding this issue https://github.com/react-native-community/react-native-tab-view/issues/716

### Test plan

Added a new example FastIndexChange. The screen is active if it's pink, before this change it took around 1s before the new screen got pink. Now it's instant.

### DIFF
--- a/src/Pager.js
+++ b/src/Pager.js
@@ -284,6 +284,11 @@ export default class Pager<T: Route> extends React.Component<Props<T>> {
         set(state.finished, FALSE),
         set(this._index, index),
         startClock(this._clock),
+        call([this._index], ([value]) => {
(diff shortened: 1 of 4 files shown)


putuyoga:
Now it looks good on my app. I will use this branch on my `packages.json`, instead official one. Thanks @lindesvard

src/Pager.js
```diff
@@ -284,6 +284,11 @@ export default class Pager<T: Route> extends React.Component<Props<T>> {
         set(state.finished, FALSE),
         set(this._index, index),
         startClock(this._clock),
+        call([this._index], ([value]) => {
```
osdnk:
```suggestion
        onChange(call([this._index], ([value]) => {
```

satya164:
Closing in favor of #764

PR #722, opened 2019-03-21
Support horizontal bounces for PagerScoll to make it looks like iOS component

azurechen:
### Motivation

Currently the Pager bounce is disabled and it looks like an Android Pager not iOS scrollview.
Support horizontal bounces for `PagerScoll` to make it looks like an iOS component.

`bouncesEnabled` props is default `false` to avoid affecting existing users.

This PR is for v1.0 because most developers are still using this version.
I will create another PR for v2.0 as soon as possible.

ezgif com-video-to-gif (10)

### DIFF
--- a/src/PagerScroll.js
+++ b/src/PagerScroll.js
@@ -140,7 +140,7 @@ export default class PagerScroll<T: *> extends React.Component<
       navigationState,
       onSwipeStart,
       onSwipeEnd,
-      bouncesEnabled,
+      bounces,
     } = this.props;
 
     return (
@@ -153,8 +153,8 @@ export default class PagerScroll<T: *> extends React.Component<
         overScrollMode="never"
         scrollEnabled={this.props.swipeEnabled}
         automaticallyAdjustContentInsets={false}
-        bounces={bouncesEnabled || false} // default false
-        alwaysBounceHorizontal={bouncesEnabled || false} // default false
+        bounces={bounces || false} // default false
+        alwaysBounceHorizontal={bounces || false} // default false
         scrollsToTop={false}
         showsHorizontalScrollIndicator={false}
         scrollEventThrottle={1}
--- a/src/PagerScroll.js
+++ b/src/PagerScroll.js
@@ -152,8 +152,8 @@ export default class PagerScroll<T: *> extends React.Component<
         overScrollMode="never"
         scrollEnabled={this.props.swipeEnabled}
         automaticallyAdjustContentInsets={false}
-        bounces={false}
-        alwaysBounceHorizontal={false}
+        bounces
+        alwaysBounceHorizontal
         scrollsToTop={false}
         showsHorizontalScrollIndicator={false}
         scrollEventThrottle={1}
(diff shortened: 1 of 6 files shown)


src/PagerScroll.js
```diff
@@ -140,6 +140,7 @@ export default class PagerScroll<T: *> extends React.Component<
       navigationState,
       onSwipeStart,
       onSwipeEnd,
+      bouncesEnabled,
```
satya164: Thanks for the PR. Can you rename `bouncesEnabled` to just `bounces` to be consistent with RN's ScrollView naming?

src/TypeDefinitions.js
```diff
@@ -46,6 +46,7 @@ export type PagerRendererProps<T> = PagerCommonProps<T> & {
 export type PagerCommonProps<T> = {
   animationEnabled?: boolean,
   swipeEnabled?: boolean,
+  bouncesEnabled?: boolean,
```
satya164: I think this should be in `PagerScroll`'s props because it's not a common prop. We also need to add it to the `propTypes`

src/PagerScroll.js
```diff
@@ -152,8 +153,8 @@ export default class PagerScroll<T: *> extends React.Component<
         overScrollMode="never"
         scrollEnabled={this.props.swipeEnabled}
         automaticallyAdjustContentInsets={false}
-        bounces={false}
-        alwaysBounceHorizontal={false}
+        bounces={bouncesEnabled || false} // default false
```
satya164: We can use `defaultProps` to set it to `false` instead

src/TypeDefinitions.js
azurechen: I can't only set `bounces` prop for `PagerScroll` because this prop should be assigned for `TabView`.

`TabFlow`'s type is 
```
type Props<T> = PagerCommonProps<T> &
  PagerExtraProps & {
    navigationState: NavigationState<T>,
    onIndexChange: (index: number) => mixed,
    initialLayout?: Layout,
    renderPager: (props: *) => React.Node,
    renderScene: (props: SceneRendererProps<T> & Scene<T>) => React.Node,
    renderTabBar: (props: SceneRendererProps<T>) => React.Node,
    tabBarPosition: 'top' | 'bottom',
    useNativeDriver?: boolean,
    style?: ViewStyleProp,
  };
```
And `PagerScoll`'s type is 
```
type Props<T> = PagerRendererProps<T>

export type PagerRendererProps<T> = PagerCommonProps<T> & {
  layout: Layout & {
    measured: boolean,
  },
  navigationState: NavigationState<T>,
  panX: Animated.Value,
  offsetX: Animated.Value,
  jumpTo: (key: string) => mixed,
  useNativeDriver: boolean,
  children: Node,
};
```

The shared type of `TabView` and `PagerScroll` is `PagerCommonProps`. I can only set `bounces` type in `PagerCommonProps` to pass flowtype validation.

But Android user will not be affect by this prop. And I have added `bounces` prop to `PropTypes` file.

src/PagerScroll.js
azurechen: nice. It is better.

src/PagerScroll.js
azurechen: Sure. `bounces` make sense.

azurechen:
hi @satya164 . should I need to update the branch? And is this PR ready to be merged?

PR #740, opened 2019-04-10
Add rtl support for swipe

aliazizi:
Change pager.js swipe controll for rtl support

### Motivation

Enable swipping for rtl language

### DIFF
--- a/src/Pager.js
+++ b/src/Pager.js
@@ -388,7 +388,12 @@ export default class Pager<T: Route> extends React.Component<Props<T>> {
           set(this._offsetX, this._position),
         ]),
         // Update position with previous offset + gesture distance
-        set(this._position, I18nManager.isRTL ? sub(this._offsetX, this._gestureX) : add(this._offsetX, this._gestureX)),
+        set(
+          this._position,
+          I18nManager.isRTL
+            ? sub(this._offsetX, this._gestureX)
+            : add(this._offsetX, this._gestureX)
+        ),
         // Stop animations while we're dragging
         stopClock(this._clock),
       ],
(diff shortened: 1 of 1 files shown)


satya164:
This is huge!

Thanks for the PR. I tested it and it works. There's a small issue where the screen shakes when you touch it or when you lift your finger.

alikazemkhanloo:
The issue that @satya164 just mentioned is because of  this line
https://github.com/react-native-community/react-native-tab-view/pull/740/files#diff-13c90f71216108ed4d0e57b27084346bL291

I used  `timing` and its configs from its next line instead of `spring` , and it worked. But this is not a solution. just a clue to solve the problem.

aliazizi:
Ok, I spent many hours to solve this issue and I'm stuck in it.

I figure out I should multiply this._velocityX to -1 in RTL mode for spring animation, but it's get me an error for prop type require. I spent plenty time to solve this but I'm stuck in it.

also I figure out the index calculated and sent to translateX ( in both rtl and ltr ) calculate multi time and for example when I swipe from index 0 to index 1, base on velocity Its maybe plus two time and the calculate value will be 2 in some time.

osdnk:
or maybe not 🤔

satya164:
@aliazizi we'll look into it. thanks again for the PR :)

src/Pager.js
```diff
@@ -189,6 +190,10 @@ export default class Pager<T: Route> extends React.Component<Props<T>> {
   _swipeDistanceThreshold = new Value(this.props.swipeDistanceThreshold || 180);
   _swipeVelocityThreshold = new Value(this.props.swipeVelocityThreshold);
 
+  // Velocity of a pan movement used for a natural swipe
+  // Setting before gesture
+  _initialVelocityForSpring = new Value(0);
```
satya164: Looks good. Will be great if you can mention why we need to do it

aliazizi:
Your welcome,
I see the commit and it's worked,
Thanks for improving

PR #747, opened 2019-04-18
feat: export extra animation options for Pager

osdnk:
### Motivation

Sometimes I suppose a user might want to define own behavior of timing of spring and I want to make it possible with this PR

### Test plan

Add some config (e.g. `extraSpringConfig`) to `Pager`.

### DIFF
--- a/src/Pager.js
+++ b/src/Pager.js
@@ -79,6 +79,20 @@ const DIRECTION_RIGHT = -1;
 const SWIPE_DISTANCE_MINIMUM = 20;
 const SWIPE_DISTANCE_MULTIPLIER = 1 / 1.75;
 
+const SPRING_CONFIG = {
+  damping: 30,
+  mass: 1,
+  stiffness: 200,
+  overshootClamping: true,
+  restSpeedThreshold: 0.001,
+  restDisplacementThreshold: 0.001,
+};
+
+const TIMING_CONFIG = {
+  duration: 250,
+  easing: Easing.out(Easing.cubic),
+};
+
 export default class Pager<T: Route> extends React.Component<Props<T>> {
   static defaultProps = {
     swipeVelocityThreshold: 1200,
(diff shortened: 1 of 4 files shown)


src/Pager.js
```diff
@@ -96,6 +96,8 @@ const TIMING_CONFIG = {
 export default class Pager<T: Route> extends React.Component<Props<T>> {
   static defaultProps = {
     swipeVelocityThreshold: 1200,
+    extraSpringConfig: {},
```
satya164: Let's rename these to `springConfig` and `timingConfig` and then make them updatable when you change props.

src/TabView.js
```diff
@@ -94,6 +94,8 @@ export default class TabView<T: Route> extends React.Component<
       swipeEnabled,
       swipeDistanceThreshold,
       swipeVelocityThreshold,
+      timingConfig,
```
osdnk: 👀

src/types.js
```diff
@@ -41,4 +41,14 @@ export type PagerCommonProps = {|
   swipeVelocityThreshold?: number,
   onSwipeStart?: () => mixed,
   onSwipeEnd?: () => mixed,
+  springConfig: {|
```
osdnk: I think it's possible to reanimated nodes here

osdnk: No, but not now xD

src/Pager.js
```diff
@@ -190,6 +202,45 @@ export default class Pager<T: Route> extends React.Component<Props<T>> {
...
+      this.props.springConfig.restDisplacementThreshold !== undefined
+        ? this.props.springConfig.restDisplacementThreshold
+        : 0.001
+    ),
+    overshootClamping: true,
+  };
+
+  _timingConfig = {
+    duration: new Value(
+      typeof this.props.timingConfig.duration !== 'undefined'
```
osdnk: @satya164

PR #752, opened 2019-04-23
fix: remove tabBar shadow

Taylor123:
### Motivation

There's a default shadow on the parent `Animated.View` which caused a shadow on the text rendered from the `renderLabel` prop. In my opinion, there should be no shadow by default and can be added during implementation.

### Test plan

trivial change


satya164:
The library follows material design for the default styling. If you don't want a shadow in your design, you can remove it.

PR #760, opened 2019-04-30
docs: improve docs regarding performance

satya164:


README.md
```diff
@@ -540,6 +555,14 @@ renderScene = ({ route }) => {
 
 Nesting the `TabView` inside a vertical `ScrollView` will disable the optimizations in the `FlatList` components rendered inside the `TabView`. So avoid doing it if possible.
 
+### Use `lazy` and `lazyPlaceholder` props to render routes as needed
```
osdnk: Idk but maybe we can repeat here if this option is enabled by default

satya164: @osdnk it's disabled by default 😛

satya164: Done

Issue #743, opened 2019-04-13
Jumping to tab programmatically not working in iOS


anirudhramanan:
### Current behaviour

On changing the tab index programmatically, it only changes the tab indicator but not the scene below. The scene still renders the first screen, but the tab indicator points to the current index.

### Expected behaviour

The scene should be updated as well.

### Screenshots (if applicable)

<img width="238" alt="Screenshot 2019-04-13 at 1 34 14 PM" src="https://user-images.githubusercontent.com/16556984/56076591-eb233c80-5df0-11e9-8deb-dbcbff7e2651.png">

### Your Environment

iOS

wizebin:
I can't seem to recreate your issue, the code in this branch works:

https://github.com/wizebin/TabViewVelocityIssue/tree/programmatic-jumping

You could try forking this repo and demonstrating your issue.

Buddikazz:
i also canot move tab programitically with using index in IOS.android work fine.

manurana:
I am also seeing a weird issue with the component not respecting a programmatic tab change only on a iOS device. An iOS emulator, and android (device/emulator) work fine.
Don't know if its related. Will try to set up an example to demonstrate

gamingumar:
Are you using the onIndexChange callback to change your index?

anirudhramanan:
To repro the issue, set the index in the initialState (in the constructor) to second or third tab. The onIndexChange callback is working though.

sienkashing:
I had this issue when the version of react-native-tab-view I was using was v1.3.2.

Upgrading it to v2.2.0 manage to fix the issue for me. (You will need to install [react-native-reanimated](https://github.com/kmagiera/react-native-reanimated) for this update.)

satya164:
Closing due to incomplete issue template

PR #762, opened 2019-05-02
fix: setting initial tab behaviour for iOS RN 0.59.x

levibuzolic:
Fix an issue where the initial tab is ignored on iOS and React Native 0.59.x when the initial tab is not the first tab in the set.

This uses the newly introduced [scrollToOverflowEnabled prop](https://github.com/facebook/react-native/commit/e3ac329).

It seems counterintuitive to have this along with `overScrollMode="never"` but `overScrollMode` really controls the ability to scroll past the ends of a `ScrollView`, while `scrollToOverflowEnabled` allows us to manually `scrollTo()` an arbitrary position out of bounds of the `ScrollView`.

I'm unsure of the underlying change in RN0.59.x that caused this, but it seems that the ScrollView hasn't completed layout when the `_setInitialPage()` is called, so it's unable to scroll to a non-zero x value.

Should address https://github.com/react-navigation/react-navigation/issues/5754

May also close this issue: #743

### DIFF
--- a/src/PagerScroll.js
+++ b/src/PagerScroll.js
@@ -150,6 +150,7 @@ export default class PagerScroll<T: *> extends React.Component<
         keyboardDismissMode="on-drag"
         keyboardShouldPersistTaps="always"
         overScrollMode="never"
+        scrollToOverflowEnabled
         scrollEnabled={this.props.swipeEnabled}
         automaticallyAdjustContentInsets={false}
         bounces={false}


satya164:
Thanks a lot for the PR.

cc @osdnk you had this issue

osdnk:
Cool! Thanks! @jinchung, could you test if it works for us :)

PR #763, opened 2019-05-04
Added an ability to style TabBar component through TabView

bintoll:
It was not possible to style TabBar component through TabView. Now the new props to TabView were added: tabStyle, indicatorStyle, labelStyle. And now these props: style, tabStyle, indicatorStyle, labelStyle are passed to TabBar component.

### Motivation

Was not able to style TabBar component through the TabView component.

### Test plan

Pass one of these props:  style, tabStyle, indicatorStyle, labelStyle to TabView component and you will be able to see the changes in TabBar component.

### DIFF
--- a/.flowconfig
+++ b/.flowconfig
@@ -49,6 +49,9 @@
 ; Ignore modules with errors
 .*/node_modules/@expo/.*
 
+; Ignore edit history
+.*/.history
+
 [untyped]
 .*/node_modules/expo/.*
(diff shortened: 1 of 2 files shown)


satya164:
Thanks for the pull request, but I don't want to add this API. It's few extra lines of code, but it's explicit to pass props to `TabBar` and there's no confusion on how things will work when you're using a custom tab bar.

PR #764, opened 2019-05-06
fix: try to improve the performance

### LINKED ISSUE #700: Position listener sends too many events
### Current behaviour

I've tailored the tab view to be able to have a tabbar that scrolls along as the user swipes through the pages. I'm using the position event listener to receive current position and calculate how far the scrollview (tabbar) should scroll.

It works great, however the listener sends so many events that the UI starts to lag and hangs for a second or so.

### Expected behaviour

Is there any chance you can throttle the amount of events sent by the listener?

### Code sample

```
public componentDidMount() {
    this.props.addListener('position', this.adjustScroll, 20);
  }

  public adjustScroll = (value: number) => {
    cancelAnimationFrame(this.scrollResetCallback);

    if (this.scrollView) {
      this.scrollView.scrollTo({
        animated: false,
        x: this.calculateScrollValue(value),
      });
    }
  };

 public calculateScrollValue = (position: number) =>
    Math.max(
      interpolate(position, {
        inputRange: [0, this.props.navigationState.routes.length - 1],
        outputRange: [
          0,
          this.getTabMeasurements(this.tabMeasurements.length - 1).centerOffset ||
            this.props.layout.width,
        ],
      }),
      0,
    );

```

### Screenshots (if applicable)

/

### What have you tried

I've tried throttling the incoming events, but it seems the bridge still slows down because of the event stream.

### Your Environment

| software                     | version
| ---------------------------- | -------
| ios or android               | iOS
| react-native                 | 0.59
| react-native-tab-view        | 2.0.0-alpha.8
| react-native-gesture-handler | latest
| react-native-reanimated      | latest
| node                         | 10 LTS
| npm or yarn                  | yarn

satya164:
This commit aims to throttle the events from reanimated to avoid overloading JS thread. It changes a few things:
- Due to the throttling, we cannot get the same scroll as you swipe effect on scrollable tab bar, which is probably better as this didn't work smoothly on lower end devices
- We now call onIndexChange as soon as you lift your finger rather than waiting for the animation to finish, which seems to work okay
- Due to throttling, empty page for lazy loaded tabs is more pronounced, but this can be work around by providing a placeholder component

Closes #700, #716

### DIFF
--- a/src/Pager.js
+++ b/src/Pager.js
@@ -55,6 +55,7 @@ const {
   greaterThan,
   max,
   min,
+  modulo,
   multiply,
   neq,
   or,
@@ -92,6 +93,8 @@ const TIMING_CONFIG = {
   duration: 250,
   easing: Easing.out(Easing.cubic),
 };
+
+const THROTTLE_FACTOR = 16;
 
 export default class Pager<T: Route> extends React.Component<Props<T>> {
   static defaultProps = {
@@ -202,6 +205,8 @@ export default class Pager<T: Route> extends React.Component<Props<T>> {
   _gestureX = new Value(0);
   _gestureState = new Value(State.UNDETERMINED);
   _offsetX = new Value(0);
+
+  _throttler = new Animated.Value(0);
 
   // Current position of the page (translateX value)
   _position = new Value(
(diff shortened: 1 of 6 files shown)


satya164:
@osdnk do you want to review this PR?

osdnk:
I’ll see it soon

src/Pager.js
```diff
@@ -435,18 +426,37 @@ export default class Pager<T: Route> extends React.Component<Props<T>> {
...
-      onChange(
-        this._position,
-        call([this._position], this._handlePositionChange)
-      )
+      onChange(this._position, [
+        cond(
+          eq(this._throttler, 0),
+          call([this._position], this._handlePositionChange)
+        ),
+        set(this._throttler, modulo(add(this._throttler, 1), THROTTLE_FACTOR)),
```
osdnk: Maybe we want to set `this._throttler` to zero the beginning to make it sure that listener will be invoked at least once at the gesture.

satya164: Do you mean I should set it to 0 when the gesture starts?

PR #767, opened 2019-05-08
refactor: drop position listeners in favor of enter listeners

satya164:
This drops the undocumented position listener which can cause perf issues by overloading the bridge in favor of a enter listener which can be used for lazy loading

### DIFF
--- a/src/Pager.js
+++ b/src/Pager.js
@@ -11,6 +11,7 @@ import type {
   Route,
   Listener,
   PagerCommonProps,
+  EventEmitterProps,
 } from './types';
 
 type Props<T: Route> = {|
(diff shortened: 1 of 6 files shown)


osdnk:
Is it actually needed and has a big performance impact?

satya164:
@osdnk After throttling there shouldn't be much issues with perf, but:

1. As you pointed out, throttling isn't great for all use cases, you might still want non-throttled updates
2. The `enter` listener makes the lazy load experience faster (similar to how it was with no throttling)

I think for the use cases where someone needs position updates, maybe it's better to accept an `Animated.Value` as a prop which we update, so they can use a native animated value instead of JS based listener. And in the cases where they do need a JS based listener, they can add a listener to the animated value they are passing.

PR #768, opened 2019-05-08
fix: tweak spring config for animation

satya164:
This is configuration used by stack navigator. Feels very similar to what reddit app has.


satya164:
@osdnk can you check if this feels "natural" on iOS?

PR #772, opened 2019-05-13
feat: Add method argument to onIndexChange callback

Jyrno42:
This argument is either gesture/tap and allows one to differentiate between the action user took to change the active scene.

### Motivation

In the app I am building, I need to track the change method for analytics.

### Test plan

Verified manually that this change did not cause any regressions.

### DIFF
--- a/README.md
+++ b/README.md
@@ -154,7 +154,7 @@ Example:
 
 ##### `onIndexChange` (`required`)
 
-Callback which is called on tab change, receives the index of the new tab as argument.
+Callback which is called on tab change, receives the index of the new tab as its first argument, the second argument is the change method (gesture/tap).
 The navigation state needs to be updated when it's called, otherwise the change is dropped.
 
 ##### `renderScene` (`required`)
--- a/src/Pager.tsx
+++ b/src/Pager.tsx
@@ -16,7 +16,7 @@ import {
 type Binary = 0 | 1;
 
 type Props<T extends Route> = PagerCommonProps & {
-  onIndexChange: (index: number) => void;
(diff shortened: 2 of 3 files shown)


Jyrno42:
Rebased on top of the TS conversion.

satya164:
I'm not sure about this. `jumpToIndex` being called doesn't mean it's a tab press.

> In the app I am building, I need to track the change method for analytics.

You can use the `onTabPress` callback to check tab changes due to tab press.

Jyrno42:
> I'm not sure about this. `jumpToIndex` being called doesn't mean it's a tab press.

Did you mean to write `jumpTo`? Since in the PR `jumpToIndex` of TabView only forwards the passed argument.

Using onTabPress requires me to use an instance variable to prevent sending a second analytics event from `onIndexChange` callback.

Jyrno42:
What if I change the implementation like this:

1. Add second argument to Pager.jumpTo and pass it along to calls to `this.props.onIndexChange` in its body.
2. Modify `onPress` handler of `TabBarItem` inside `TabBar.tsx` to and add a second argument 'tap' to `this.props.jumpTo` call
3. [Optional] Change constant string 'gesture' to 'other' for semantic reasons (it might be a gesture or a reaction to a changed index in props)

satya164:
I need to think about it a bit.

> Using onTabPress requires me to use an instance variable to prevent sending a second analytics event from onIndexChange callback which is not very elegant.

For now this seems reasonable. We do similar in react-navigation.

Jyrno42:
Used `onTabPress` and got the functionality I wanted.

this PR can be closed from my perspective

Issue #720, opened 2019-03-19
Large number of routes massively drop JS frame rate when switching tabs


tmmgn:
### Current behaviour

TabBar with scroll enabled is laggy and slow on android when using gestures or click on tab even if i render only <View/> inside tab

### Expected behaviour

Expect TabBar with scroll enabled will run smooth as in example

### Code sample

I have used code sample from example:
https://github.com/react-native-community/react-native-tab-view/blob/master/example/src/ScrollableTabBarExample.js

### Screenshots (if applicable)

![lags](https://user-images.githubusercontent.com/11463030/54597020-d8625580-4a46-11e9-831a-05056be9a762.gif)

### What have you tried

Have tried all the optimization tips from the documentation, nothing has chaned. Tried to reduce the number of routes, but lags is still present even with 3 routes

### Your Environment

| software                     | version
| ---------------------------- | -------
| android               |9
| react-native                 |0.59.1
| react-native-tab-view        |2.0.3
| react-native-gesture-handler |1.1.0
| react-native-reanimated      |1.0.0-alpha.12
| node                         |11.0.0
| npm or yarn                  |6.4.1

joshperry131:
We also use this in our project and have found the same issue.

I have extracted the ScollableTabBar example and given it 8 tabs instead of 4 to make the behavior obvious in the following gif:
![tabsslow](https://user-images.githubusercontent.com/24830987/54820280-80219280-4c96-11e9-8736-95d4440a4376.gif)

The indicator for which tab is selected lags behind the selected tab buy quite a bit if navigating quickly to a tab a few indexes along via gesture. The scrolling behavior of the tabBar is necessary for our application.

tmmgn:
I used this library as temporary workaround for scrollable tabs - https://github.com/ptomasroos/react-native-scrollable-tab-view. Looks pretty fast, give it a try if you need to hurry with your project

xaviershay:
I think this might be a dupe of #712

xaviershay:
I did what @joshperry131 did above, but pushed it up if anyone else wants to try: https://github.com/xaviershay/react-native-tab-view/tree/720-repro

The problem scales with number of tabs, so drastically increasing the number makes it exceedingly obvious.

xaviershay:
I think I've narrowed this down, but I'm not sure how to fix yet.

[...]

PR #773, opened 2019-05-14
fix: memoize animated nodes to improve performance

satya164:
Currently the animated nodes used for styles are re-created in render, which means that reanimated does a lot of expensive work detaching and reattaching them. Due to this, the framerate of the JS thread goes down to 0 FPS even on high-end devices when switching tabs quickly.

This PR implements memoization for the animated nodes so they are only recreated when necessary. On low-end devices, this keeps the JS frame rate around 12+ FPS on low-end devices and 40+ on high-end devices.

Related to #720


RobertTownley:
Thanks so much for this change; the app is running buttery-smooth now. You're a life saver

PR #810, opened 2019-06-19
fix: don't consider velocity and gesture distance if gesture didn't end

### LINKED ISSUE #806: TabView seems to ignore state if keyboard was opened
# Bug

`TabView` seems to be ignoring state changes - or tab taps - when some input in the scene receives focus

## Environment info

React native info output:

```bash
React Native Environment Info:
    System:
      OS: Linux 5.0 Fedora 29 (Workstation Edition) 29 (Workstation Edition)
      CPU: (4) x64 Intel(R) Core(TM) i3-3220 CPU @ 3.30GHz
      Memory: 524.91 MB / 7.66 GB
      Shell: 5.6.2 - /bin/zsh
    Binaries:
      Node: 11.12.0 - /tmp/fnm-shell-4858630/bin/node
      Yarn: 1.15.2 - /tmp/fnm-shell-4858630/bin/yarn
      npm: 6.9.0 - /tmp/fnm-shell-4858630/bin/npm
    SDKs:
      Android SDK:
        API Levels: 25, 27, 28
        Build Tools: 27.0.3, 28.0.3
        System Images: android-28 | Google Play Intel x86 Atom
    npmPackages:
      react: 16.8.3 => 16.8.3 
      react-native: 0.59.2 => 0.59.2 
    npmGlobalPackages:
      react-native-cli: 2.0.1
      react-native: 0.59.2
```

```
"react-native-gesture-handler": "1.1.0",
"react-native-reanimated": "1.0.1",
"react-native-tab-view": "2.6.1",
```

## Steps To Reproduce

The code below may help you to reproduce this, I'm not sure if this is caused by the use of react hooks, but when I click on any tab after a input focus, the state is set correctly (`console.log` shows that the state is set) but the tab view doesn't show the scene for the current state.
If I swipe to another scene and then click on any tab everything works fine again, if I focus on any input, the problem appears again

Describe what you expected to happen:

1. I type something in any input on any scene, then click on any tab
2. The tab view shows the correct scene based on the current state

## Reproducible sample code

Running without debugger in a real device (One plus 5T):

```typescript
import React, { useState } from 'react'
import { TabView } from 'react-native-tab-view'
import { Dimensions, Text, TextInput, View } from 'react-native'

export default () => {
    const [currentIndex, setCurrentIndex] = useState(0)
    const routes = [
        { key: 'dataTab', title: 'Dados' },
        { key: 'followUpTab', title: 'Follow up' }
    ]

    const renderScene = ({ route }) => {
        return (
            <View style={ { flex: 1 } }>
                <Text> { route.key } </Text>
                <TextInput />
            </View>
        )
    }

    return (
        <TabView
            navigationState={ {
                index: currentIndex,
                routes
            } }
            renderScene={ renderScene }
            onIndexChange={ setCurrentIndex }
            initialLayout={ { width: Dimensions.get('window').width } }
        />
    )
}

```

### LINKED ISSUE #809: tabs don't switch after a vertical gesture
# Bug
A "Quick start" code freezes on the selected tab after a vertical gesture on a TabView. The code there uses regular View elements but the same thing happens when you have scrollable components as a child.

## Environment info
Device: Every Android Device (afaik)

Library version: ^2.6.0

## Steps To Reproduce
Play with tabs. Switch with buttons. Switch with gestures. So far it works. Now make a vertical gesture like scrolling. The selected tab should freeze, buttons do not do anything until you change the tab by swiping.

## Reproducible sample code
https://snack.expo.io/@beqa/tab-view-example-not-working

satya164:
previously, when gesture wasn't active, we checked if the gesture exceeded the minimum threshold (this was due to activeOffsetX not being respected on Android),
- if yes, we calculate the next position based on gesture distance or velocity
- if no, we transition to what the value of index was

this addressed 2 cases: the first condition would trigger when a gesture was performed, the second one will trigger if the index changed due to other means such as state update.
this check was mistakenly removed in e7f832c

the new code checked for `State.END`, which resulted in the tab switch due to state update not working at times. this is because the `gestureState` value is different when a gesture was cancelled, e.g. due to a vertical gesture or focus on an input element.

closes #806, closes #809

### DIFF
--- a/package.json
+++ b/package.json
@@ -16,7 +16,7 @@
     "lint": "eslint --ext .js,.ts,.tsx .",
     "release": "release-it",
     "example": "yarn --cwd example",
-    "bootstrap": "yarn && yarn example",
+    "bootstrap": "yarn example && yarn",
     "prepare": "bob build"
   },
   "publishConfig": {
--- a/src/Pager.tsx
+++ b/src/Pager.tsx
@@ -42,6 +42,7 @@ const {
   Clock,
   Value,
   onChange,
+  and,
   abs,
   add,
   block,
@@ -188,7 +189,7 @@ export default class Pager<T extends Route> extends React.Component<Props<T>> {
   // Current state of the gesture
   private velocityX = new Value(0);
   private gestureX = new Value(0);
-  private gestureState = new Value(State.END);
+  private gestureState = new Value(State.UNDETERMINED);
   private offsetX = new Value(0);
 
   // Current progress of the page (translateX value)
@@ -415,6 +416,7 @@ export default class Pager<T extends Route> extends React.Component<Props<T>> {
     divide(abs(this.velocityX), this.velocityX),
     0
   );
+
   private extrapolatedPosition = add(
     this.gestureX,
     multiply(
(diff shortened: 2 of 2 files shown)


osdnk:
Can someone make a release? My internet connection here is too slow to install `yarn` 😒

satya164:
> Terribly sorry for breaking changes and these issues.

Nah. I should've written some comments initially on why that was done. Easy mistake to make 😃

PR #751, opened 2019-04-23
feat: add dynamic tab width support

Taylor123:
### Motivation

There is currently no support for tabs that have a variable width. This is a common design and seems to be something the community wants  (https://github.com/react-native-community/react-native-tab-view/issues/356).

Credit to @jaitd for the initial related PR (https://github.com/react-native-community/react-native-tab-view/pull/544)

### Test plan

Checked with the example app and tested on my production app. I can publish it to npm to create a snack if it's necessary.

PR #823, opened 2019-06-25
Fix dynamic tab perf

ankeetmaini:
This works on top of #751 

I tried multiple things to improve the performance, but I think the main issue was multiple calls to `setState` in `onLayout` of `<TabBarItem/>`. This PR only sets state once when all `onLayout`s have been fired by using a private variable which is cleared later. I could see visible perf improvement and the initial scroll animation of `TabBarItem`s happens in a jiffy!

I earlier tried to delay rendering the indicator as well as tab bar items after the current tab but it didn't have any visible improv.

Please see 06f66d1 for changes.

PS: I'm totally new to react-native and also haven't used this lib before, so I'm not privy to the internal workings. Let me know if this approach doesn't abide by the theme/structure of this codebase!

cc: @Taylor123 @satya164


src/TabBar.tsx
```diff
@@ -155,7 +178,41 @@ export default class TabBar<T extends Route> extends React.Component<
...
+        .reduce(
+          (acc: number, tabWidth: number) => (acc += tabWidth),
+          tabWidths[targetIndex] * 0.5
+        );
+
+      let diff =
+        tabWidths[targetIndex] * 0.5 +
+        tabWidths[props.navigationState.index] * 0.5;
+
+      if (i > props.navigationState.index) {
```
satya164: I'm a bit confused about why the current navigation state matters when scrolling to an index. This function should scroll to a specific index without caring about current state.

Though this function is triggered when the index changes, so `i` is the same as the current index anyway.

src/TabBar.tsx
```diff
@@ -155,7 +178,41 @@ export default class TabBar<T extends Route> extends React.Component<
...
-    const { layout } = state;
+    const { layout, tabWidths } = state;
+
+    if (props.dynamicWidth) {
+      const targetIndex =
+        i > props.navigationState.index &&
+        i < props.navigationState.routes.length - 1
+          ? Math.ceil(i)
+          : Math.floor(i);
+      let dynamicAmount = tabWidths
```
satya164: Can you change the variable name to be more descriptive? I assume this means the scroll amount?

src/TabBarItem.tsx
```diff
@@ -45,7 +48,7 @@ type Props<T extends Route> = {
 const DEFAULT_ACTIVE_COLOR = 'rgba(255, 255, 255, 1)';
 const DEFAULT_INACTIVE_COLOR = 'rgba(255, 255, 255, 0.7)';
 
-export default class TabBarItem<T extends Route> extends React.Component<
+export default class TabBarItem<T extends Route> extends React.PureComponent<
```
satya164: Is changing to `PureComponent` needed here?

src/TabBarItem.tsx
```diff
@@ -63,21 +66,24 @@ export default class TabBarItem<T extends Route> extends React.Component<
     }
   );
 
-  private getInactiveOpacity = memoize((position, routes, tabIndex) => {
```
satya164: Looks like this method didn't change. Can we revert this part of the diff to what it was?

src/TabBarItem.tsx
```diff
@@ -202,7 +209,7 @@ export default class TabBarItem<T extends Route> extends React.Component<
       scrollEnabled === true;
 
     const tabContainerStyle: ViewStyle = {};
-    const itemStyle = isWidthSet ? { width: tabWidth } : null;
+    const itemStyle = isWidthSet && !dynamicWidth ? { width: tabWidth } : null;
```
satya164: We should use `width: 'auto'` in `tabStyle` for this

src/TabBarIndicator.tsx
```diff
@@ -6,39 +6,107 @@ import memoize from './memoize';
 import { Route, SceneRendererProps, NavigationState } from './types';
 
 export type Props<T extends Route> = SceneRendererProps & {
+  dynamicWidth?: boolean;
```
satya164: I'm not a fan of the indicator being aware of this prop. Maybe a `getTabWidth` prop which accepts an index and returns a width is cleaner. It could return the measured width or pre-calculated (layout.width / routes.length) depending on if the widths are dynamic

src/TabBarIndicator.tsx
```diff
@@ -6,39 +6,107 @@ import memoize from './memoize';
...
+      routes: Route[],
+      tabWidths: number[],
+      width: number,
+      dynamicWidth?: boolean
+    ) => {
+      if (dynamicWidth) {
+        let inputRange: number[] = [];
+        let outputRange: number[] = [];
+        let totalWidth = 0;
+        routes.forEach((_route: Route, i: number) => {
```
satya164: Instead of a `forEach`, I feel a `map` for `inputRange` and a `reduce` for `outputRange` is cleaner

src/TabBarIndicator.tsx
```diff
@@ -6,39 +6,107 @@ import memoize from './memoize';
...
+    (
+      position: Animated.Node<number>,
+      routes: Route[],
+      tabWidths: number[],
+      dynamicWidth?: boolean
+    ) => {
+      if (dynamicWidth) {
+        let inputRange: number[] = [];
+        let outputRange: number[] = [];
+        routes.forEach((_route: Route, i: number) => {
```
satya164: I think `map` is cleaner here

src/TabBarIndicator.tsx
```diff
@@ -6,39 +6,107 @@ import memoize from './memoize';
...
+          inputRange,
+          outputRange,
+        });
+      }
+      return 1;
+    }
+  );
+
+  private getStyle = memoize((routes: Route[], dynamicWidth?: boolean) =>
+    dynamicWidth ? { width: 1 } : { width: `${100 / routes.length}%` }
```
satya164: Probably worth adding a comment here, coz `width: 1` isn't very obvious.

src/TabBar.tsx
```diff
@@ -291,9 +351,22 @@ export default class TabBar<T extends Route> extends React.Component<
...
-            {routes.map((route: T) => (
+            {routes.map((route: T, i: number) => (
               <TabBarItem
+                onLayout={({ nativeEvent: { layout } }) => {
+                  if (!dynamicWidth) return;
+
+                  this._actualTabWidths[i] = layout.width || tabWidth;
+
+                  // check if this is the last onLayout call
+                  if (!allDone(this._actualTabWidths, routes.length)) return;
```
satya164: I'm curious on how this works if one of the tab's width changes/number of tabs change

src/TabBar.tsx
```diff
@@ -59,6 +60,13 @@ export type Props<T extends Route> = SceneRendererProps & {
 
 type State = {
   layout: Layout;
+  tabWidths: number[];
+};
+
+const allDone = (arr: number[], total: number) => {
+  return Array.from({ length: total }).reduce<boolean>((result, _, i) => {
```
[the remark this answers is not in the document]
ankeetmaini: yes!

ankeetmaini:
Most of it is done by @Taylor123 I'll try to answer as much as I can!

RPlaha0390:
This works straight out of the box! Great work @ankeetmaini. I have forked this PR and run it on a iOS  simulator. It works perfectly and very smooth. The Indicator grows and shrinks seamlessly when switching tabs. Amazing! Address the review @satya164 has requested please so we can get this merged into `master`.

src/TabBar.tsx
```diff
@@ -155,7 +178,41 @@ export default class TabBar<T extends Route> extends React.Component<
...
   private getScrollAmount = (props: Props<T>, state: State, i: number) => {
-    const { layout } = state;
+    const { layout, tabWidths } = state;
+
+    if (props.dynamicWidth) {
+      const targetIndex =
+        i > props.navigationState.index &&
+        i < props.navigationState.routes.length - 1
+          ? Math.ceil(i)
+          : Math.floor(i);
```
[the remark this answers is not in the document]
Taylor123: This is left over from the older implementation before i rebased on master recently. So this isn't necessary anymore

src/TabBar.tsx
```diff
@@ -155,7 +178,41 @@ export default class TabBar<T extends Route> extends React.Component<
   };
 
   private getScrollAmount = (props: Props<T>, state: State, i: number) => {
-    const { layout } = state;
+    const { layout, tabWidths } = state;
+
+    if (props.dynamicWidth) {
+      const targetIndex =
```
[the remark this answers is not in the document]
Taylor123: will update and add a comment

Taylor123:
@ankeetmaini i cleaned up `getScrollAmount` on this branch if you want to pull in that latest commit to this PR https://github.com/Taylor123/react-native-tab-view/tree/fix-dynamic-tab-perf

It would let me submit a PR to your branch even though i pulled from your repo

src/TabBar.tsx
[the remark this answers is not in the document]
ankeetmaini: I am afraid, this will not be correct. Because let's say 2nd tab's onLayout fires first, then length will be 2 but the array has only one element, this will give us false positives. I've changed this to a regular for-loop to make it more readable.

src/TabBarItem.tsx
ankeetmaini: I've reverted this change, although it added types to the main function params

src/TabBarIndicator.tsx
ankeetmaini: done

src/TabBarIndicator.tsx
ankeetmaini: done

src/TabBar.tsx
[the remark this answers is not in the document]
satya164: that's what `widths.every(Boolean)` is for

src/TabBarItem.tsx
ankeetmaini: reverted

src/TabBarItem.tsx
ankeetmaini: can null be totally removed in favour of `{width: 'auto'}`?

I've totally removed `dynamicWidth` prop here and it seems to work.

ankeetmaini:
```suggestion
const itemStyle = isWidthSet ? { width: tabWidth } : { width: 'auto' };
```

src/TabBar.tsx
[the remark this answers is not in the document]
ankeetmaini:
```js
const widths = [];
// assuming third tab's onLayout fires first
widths[3] = 24
// now widths is an array but has only one item present at index=3
// but JS  => widths.length = 4 and it will satisfy widths.length === length
// and it'll satisfy widths.every(Boolean) as every will run only for one item present at index=3 and not for index=0,1,2


```
widths.every(Boolean)
true

So we can't rely on this as it'll give false positive everytime.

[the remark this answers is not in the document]
satya164: Hmm, I think we just need an `Array.from` to convert the holes to proper array:

```js
widths.length === total && Array.from(widths).every(Boolean)
```

src/TabBar.tsx
```diff
@@ -291,8 +336,22 @@ export default class TabBar<T extends Route> extends React.Component<
...
+                  this.actualTabWidths[i] = layout.width || tabWidth;
+
+                  // check if this is the last onLayout call
+                  if (!didAllOnLayoutsFire(this.actualTabWidths, routes.length))
+                    return;
+
+                  // all onLayout's have been fired
+                  this.setState({ tabWidths: [...this.actualTabWidths] });
+                  // clear
+                  this.actualTabWidths = [];
```
satya164: I think we should leave this array as is. What I'm concerned about is that after initial measurement of all tabs, say later one tab label at index 3 updates, changing its width. Now `onLayout` will fire only for that tab, but we will never update the state because all other indexes will have holes in the `actualTabWidths` array.

If we leave this array as is, next time even if few of the tabs change their width, our condition for checking if all layouts have been measured will still be true and `setState` will fire correctly.

Worst case is, if multiple tabs are updating, it'll cause multiple setStates, but it's better than having incorrect state.

ankeetmaini:
@satya164 I've seen this, will work on this and update (this weekend, mostly).

satya164:
@ankeetmaini Thank you. Take your time :)

ankeetmaini:
@satya164 I think I'm almost there! I've incorporated all the reviews removed `dynamicWidth` prop altogether. 

I have two doubts, in `getTabwidth` in `TabBar.tsx` I am not sure what happens if scroll is enabled?
Also I animated the width property but it just blips while animating. I am not sure how to animate it smoothly. Can you help?

src/TabBarIndicator.tsx
```diff
@@ -6,39 +6,107 @@ import memoize from './memoize';
...
         style={[
           styles.indicator,
-          { width: `${100 / routes.length}%` },
+          dynamicStyle,
           // If layout is not available, use `left` property for positioning the indicator
           // This avoids rendering delay until we are able to calculate translateX
           width
-            ? { transform: [{ translateX }] as any }
+            ? { transform: [{ translateX }, { scaleX }] as any }
             : { left: `${(100 / routes.length) * navigationState.index}%` },
```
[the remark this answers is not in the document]
ankeetmaini: I am not sure about what's this.

satya164:
I'll try to do it tomorrow.

satya164:
@ankeetmaini

> I have two doubts, in `getTabwidth` in `TabBar.tsx` I am not sure what happens if scroll is enabled?

We don't need separate logic for that. The previous logic calculated a width so that half of a tab would be visible only when `scrollEnabled` is `true`. With `width: 'auto'`, we don't need to calculate custom width.

> Also I animated the width property of `TabBarIndicator` but it just blips while animating. I am not sure how to animate it smoothly.

This was only on Android, right? I think this was because the animated node was getting recreated every render. I have fixed the memoization to avoid it and it seems to animate smoothly now.

I have made some improvements in https://github.com/react-native-community/react-native-tab-view/pull/837. You can see what in the PR description.

The only thing I'm not satisfied about is the logic determining how much to scroll the tab. Currently it doesn't center the active item (for example, in the example, the chat tab item doesn't get centered even if there's scrollable space available). I'll try to fix it when I get time, but if you want to help with it, it'll be awesome too.

satya164:
Ok, I have updated my branch. This should fix the logic for determining center distance (it was an off-by-one error XD). Take a look when you have time :)

satya164:
Closing in favor of #823

andrestone:
This is #823 :(

sannair96:
> This is #823 :(

It's #837

PR #826, opened 2019-06-27
Add indicatorContainerStyle

stefanocdn:
Add the ability to customize the indicatorContainer so that we can achieve that kind of style.

image


satya164:
Can you update the docs?

umarrasheedmebest:
image

Container style showing on top of tabs, I want it to bottom. The container should be at the bottom. Can you please check. Thanks

PR #837, opened 2019-07-11
 feat: support `width: 'auto'` in tabStyle

satya164:
The PR is based on #751 and #823 (thanks to @Taylor123 and @ankeetmaini)

The improvements include:

1. For tab bar with `width: 'auto'`, fade in the indicator when we have the layout. This avoids sometimes noticeable jump in the indicator.
1. Memoize the `getTabWidth` prop passed to `renderIndicator`. This makes sure that the animated nodes are memoized properly as well and fixes the issue with the width jumping at end of animation on Android.
1. Make sure we adjust the scroll value only after the layout is available. This fixes an issue where sometimes the scroll position is incorrect on initial load.
1. Fixed calculation of center distance so the correct tab is centered.
1. Minor tweak to how we store tab widths in state, use an object with route key as keys which should be more accurate when new tabs are added/removed and also easier to work with.
1. Minor change which avoids adding an `onLayout` listener if tab width isn't auto.

### DIFF
--- a/README.md
+++ b/README.md
@@ -529,6 +529,8 @@ Boolean indicating whether the tab bar bounces when scrolling.
 
 Style to apply to the individual tab items in the tab bar.
 
+By default, all tab items take up the same pre-calculated width based on the width of the container. If you want them to take their original width, you can specify `width: 'auto'` in `tabStyle`.
+
 ##### `indicatorStyle`
 
 Style to apply to the active indicator.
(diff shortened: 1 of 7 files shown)


quicksilverr:
@satya164 But, I've notice a slight issue here , for the last element, when scrollEnabled is true. 

The indicator width is no adjusting according to the tab item width, and overflowing to the left. 

Screenshot_1666354568

PR #839, opened 2019-07-13
feat: add springVelocityScale effect

osdnk:
### Motivation

In this PR I add springVelocityScale parameter as a config of Pager. Toss defines how much velocity of gesture impacts on an initial velocity of Pager. Found it very useful in bottom sheet. 

Also, @christianbaroni asked me for doing it so here you are, Christian 😄 

However, while writing this code I figured out interesting bug in Pager which led to fact that initial velocity of spring was always zero. It happened bc previously velocity was set only if clock was not running. However few lines above we made similar check and if clock was not running... then we were starting it, so as the result there was no chance that in current state clock was not running. 
I fixed it by moving starting clock few lines below.

### Test plan
 You may add `springVelocityScale ={4}` param to `ScrollableTabBarExample`

### DIFF
--- a/README.md
+++ b/README.md
@@ -318,6 +318,10 @@ Configuration object for the spring animation which occurs after swiping. Suppor
 
 ##### `initialLayout`
 
+Number for determining how meaningful is gesture velocity for calculating initial velocity of spring animation. Defaults to `0`.
+
+##### `initialLayout
+
 Object containing the initial height and width of the screens. Passing this will improve the initial rendering performance. For most apps, this is a good default:
 
 ```js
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -318,6 +318,10 @@ Configuration object for the spring animation which occurs after swiping. Suppor
 
 ##### `initialLayout`
```
satya164: Typo :)

src/Pager.tsx
```diff
@@ -144,6 +148,14 @@ export default class Pager<T extends Route> extends React.Component<Props<T>> {
       );
     }
 
+    if (prevProps.springVelocityScale !== springVelocityScale) {
+      this.springVelocityScale.setValue(
+        springVelocityScale != null
```
satya164: Remove null check?

osdnk: nah, ts

Issue #880, opened 2019-09-11
Gestures don't work on parent Navigator when TabNavigators are nested


computerjazz:
Previous to react-navigation 4.x, when tab navigators were nested, a swipe on a child tab screen would navigate the child tab navigator until its final tab, then a further swipe would navigate to the next parent tab. As of 4.x, this behavior is broken, and swiping only navigates within the child/leaf navigator.

### Current behaviour

Child navigators capture all swipe gestures.

### Expected behaviour

Child navigators only capture the gesture if they can swipe, otherwise they refuse, allowing the parent to pick it up and navigate across parent tabs.

### Code sample

https://snack.expo.io/@computerjazz/tabnavswipe

### Screenshots (if applicable)

![tabbug](https://user-images.githubusercontent.com/6730148/64720841-3b75f780-d480-11e9-90e4-43c4c3fb901c.gif)

### What have you tried

I've tracked the issue down to the switchover to RNGH. In 1.4.1, there was a mechanism for the child navigator to refuse the gesture in the case that it was at the end of its routes, which allowed the parent to pick it up: https://github.com/react-native-community/react-native-tab-view/blob/v1.4.1/src/PagerPan.js#L97-L112

This is no longer the case in 4.x, so the child _always_ captures the gesture. I have confirmed that hard setting `PanGestureHandler` `enabled` to `false` on the final route allows me to swipe through parent tabs, although this is not a viable solution: https://github.com/react-native-community/react-native-tab-view/blob/master/src/Pager.tsx#L660

This seems like a potentially tricky issue with RNGH, as the `PanGestureHandler` needs to know whether it should be active, and may require `simultaneousHandlers` and/or `waitFor`, however I haven't explored that far.

Possibly related to https://github.com/react-native-community/react-native-tab-view/issues/879 

### Your Environment

```
  Expo CLI 3.0.10 environment info:
    System:
      OS: macOS 10.14.6
      Shell: 5.3 - /bin/zsh
    Binaries:
      Node: 12.9.1 - ~/.nvm/versions/node/v12.9.1/bin/node
      Yarn: 1.17.3 - ~/.yarn/bin/yarn
      npm: 6.10.2 - ~/.nvm/versions/node/v12.9.1/bin/npm
      Watchman: 4.7.0 - /usr/local/bin/watchman
    IDEs:
      Android Studio: 3.3 AI-182.5107.16.33.5264788
[...]

PR #881, opened 2019-09-17
fix: swipe between child and parent TabNavigator

computerjazz:
- Cancels Gesture when swiping past either end of routes, which allows parent `PanGestureHandler` to become the gesture handler.
- Provides context-based mechanism for child `PanGestureHandler`s to pass their refs up to the parent in order to use in `simultaneousHandlers` and `waitFor` props.

### Motivation
Fixes https://github.com/react-native-community/react-native-tab-view/issues/880

### Test plan

- Create nested tab navigator. Verify that swipes on the child, swipes on the parent, and swipes between child and parent all behave as expected.

panfix


Gokaysim:
unfortunately, I had some issue. I have added PR with yarn add then it fails when app fetches file

/react-native-tab-view/package.json` was successfully found. However, this package itself specifies a `main` module field that could not be resolved (`/react-native-tab-view/lib/module/index.js`. Indeed, none of these files exist:

computerjazz:
I think you need to build first, which generates `lib`.  
https://github.com/react-native-community/react-native-tab-view/blob/master/package.json#L20

computerjazz:
@satya164 any updates on this? I've been patching this library with these changes in order for nested tab swiping to work.

theisgroenbech:
I would love to have this merged :)

neverlist:
I need this

neverlist:
I copied this page,But it doesn't work， 
it show ‘Animated events must have event data’
I'm sorry, my English is not very good，I hope you understand what I mean

src/Pager.tsx
```diff
@@ -458,12 +507,43 @@ export default class Pager<T extends Route> extends React.Component<Props<T>> {
...
   );
 
+  private toggleEnabled = () => {
+    if (this.state.enabled)
+      this.setState({ enabled: false }, () => {
+        this.setState({ enabled: true });
+      });
+  };
+
+  // Cancel gesture if swiping back from initial tab or forward from last tab.
```
osdnk:
```suggestion
  // Cancel gesture if swiping back from the initial tab or forward from the last tab.
```

src/Pager.tsx
```diff
@@ -458,12 +507,43 @@ export default class Pager<T extends Route> extends React.Component<Props<T>> {
...
+  // Enables parent Pager to pick up the gesture, if one exists.
```
osdnk:
```suggestion
  // Enables parent Pager to pick up the gesture if one exists.
```

PR #893, opened 2019-09-27
fix: layout breaks on orientation changes

### LINKED ISSUE #877: Switching between portrait and landscape orientation UI stucks between two tabs
### Current behaviour

Especially on scrollable tab, when i'm switching between vertical and horizontal device orientation, UI stucks between two tabs but it could be unblocked by a swipe or a press on a tab item

### Expected behaviour

when i'm switching between vertical and horizontal device orientation, UI should stay at same tab 

### Code sample

For Syntax Highlighting check this [link](https://help.github.com/en/articles/creating-and-highlighting-code-blocks)

There is a repro in [example](https://github.com/react-native-community/react-native-tab-view/tree/master/example) folder

### Screenshots (if applicable)

Screenrecorder-2019-09-09-16-41-17-382
Sorry for bad quality :/

### What have you tried

Nothing for the moment

### Your Environment

| software                     | version
| ---------------------------- | -------
| android               | 28
| react-native                 | 0.59.8 (same bug with 0.60.4)
| react-native-tab-view        | 
| react-native-gesture-handler | 1.2.1
| react-native-reanimated      | 1.0.1
| node                         | 10.15.2
| npm                  | 6.9.0

deefourple:
closes #877

This is a simple PR that wraps a setState() function in a setTimeout(). Doing so prevents layout issues when changing orientation for iOS & Android.

### Motivation

This small fix will solve [Issue 877](https://github.com/react-native-community/react-native-tab-view/issues/877). I noticed that the layout would be thrown off when switching from landscape to portrait mode. 

I believe the underlying issue here is related to React Native View's [OnLayout](https://facebook.github.io/react-native/docs/view.html#onlayout) function. It says that: 

> "the new layout may not yet be reflected on the screen at the time the event is received, especially if a layout animation is in progress."

Putting a small delay on the setState function gives the Animated View enough time to calculate the layout properly. In Vue.js I would use something like [nextTick()](https://vuejs.org/v2/api/#Vue-nextTick). This is the closest solution I could find to doing something like that. 

If anybody has a more elegant solution I'm open to trying it. 

### Reproducing Steps

1) clone the app
2) `cd example/`
3) `yarn` install the dependencies
4) `yarn run start`
5) scan QR code with the expo app
5) Click the "Scrollable tab bar (auto width)
6) Be on any tab but the first
7) rotate the device (make sure your device settings allow this)
8) note the layout is broken
9) rotating back & the layout is still off


Zanchee:
In my app, this change is only fixing the tabbar, the content is still off-center.

Zanchee:
A temporary workaround solution is to force a re-render of the tab-view by setting state with some random value. Not ideal of course, but until a working solution is found, it will do.

Thebarda:
I guess you can wrap layout's setState inside setTimeout in `TabView.tsx` file. Logic should be the same, no ?

Zanchee:
I tried that, it wasn’t successful. Idk. I’m all down for a temporary solution and fix, but this doesn’t seem like an actual viable solution. It might be better to force a re-render until we can track down a better solution. I’m always wary of adding a time-out in a function.

Thebarda:
Note : This bug is reproductible only  with Scrollable view. With a simple view, it works

Zanchee:
You’re right, but I’d rather take a performance hit, that doesn’t affect the actual framework, than implement a hacking fix that doesn’t actual fix the issue fully.

Thebarda:
> You’re right, but I’d rather take a performance hit, that doesn’t affect the actual framework, than implement a hacking fix that doesn’t actual fix the issue fully.

I completly agree :)

Zanchee:
I appreciate you looking into this! If I have some time this week, I’ll try and take a peek too, but hopefully someone else will be able to chime in.

Thebarda:
That's weird ! I'm running changes made by @deefourple. The first time, the issue was not fixed with changes, but I tried to increase timeout up to 100, it solved the issue then I came back to 1 then issue still solved.

deefourple:
@Thebarda & @Zanchee would you like me to update the PR with the code above?

Zanchee:
With the code above that means the frame is now incorrect, and will continue to get one point bigger every time this function gets hit.

deefourple:
@Zanchee Correct me if I'm wrong but it doesn't increment continuously. If we get width & height from the layout event it will re-correct itself every time we change orientation. 

It could also correct itself on the second setState so the frame isn't incorrect. 

```
this.setState(
      {
        layout: {
          height: height + 1, // other value to trigger re-render
          width: width + 1, // other value to trigger re-render
        },
      },
      () => {
        this.setState({
          layout: { // 
            height - 1, 
            width - 1,
          },
        });
      }
    );
```

I do agree there must be a better solution here. I was hoping this PR would start more conversation than be a quick merge, to be honest.

Zanchee:
@deefourple You are correct, I missed the second `setState` inside of the initial `setState`. Wouldn't that mean we now have to re-render twice? I totally agree with you I was hoping more people would jump on this conversation. Perhaps @satya164 can chime in here?

For my case, until we come with a resolution, I'm adding a `key` property as a prop when I use this component and then forcing a re-render of the component by setting a new `key` value. In my case it's not a huge amount of rending that needs to occur, so I'm fine with that over adding timeouts and other things.

deefourple:
Not an elegant solution, should be re-assessed further.

PR #908, opened 2019-10-24
Added key to route components

arhmnsh:
### Motivation

Added the required key prop to components in renderScene in documentation.

### DIFF
--- a/README.md
+++ b/README.md
@@ -199,9 +199,9 @@ Callback which returns a react element to render as the page for the tab. Receiv
 renderScene = ({ route, jumpTo }) => {
   switch (route.key) {
     case 'music':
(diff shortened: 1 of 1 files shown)


satya164:
Thanks for the PR, but `key` is not required.

arhmnsh:
Hi @satya164,
When I didn't use key, the state of one tab was getting shown in another tab.

satya164:
Can you provide a repro for it? We already add a `key` internally so this shouldn't happen https://github.com/react-native-community/react-native-tab-view/blob/6198d423e6c7d618b9abd371a91615b18cc7ca4c/src/TabView.tsx#L169

PR #909, opened 2019-10-24
Fix: Layout breaks when device orientation changes

Thebarda:
### Motivation

Should solve #877 

My understading is that bug has been introduced by this commit : https://github.com/react-native-community/react-native-tab-view/commit/30216d3

On orientation changes, TabBar layout state updates his component before his own parent component pass (to his children thus TabBar) new layout prop. 

### Test plan

Just change orientation of your device / emulator

Screenrecorder-2019-10-24-16-00-33-764
Sorry for that bad quality (thx gif)


deefourple:
@satya164 Any insight into this issue then? 

Been looking for a solution and the closest guess I have is that layout is being set before the orientation has completed.

PR #941, opened 2019-12-23
feat: expose TabView and TabBar Props

Naturalclar:
### Motivation

exposed TabView Props and TabBar Props for external use.
This will help when one wants to make a customized TabView and TabBar with fixed props to use for their projects.


todorone:
@osdnk @satya164

PR #952, opened 2020-01-22
Conditionally add scrollView on iOS instead of pager

osdnk:
### Motivation

Might deliver a bit better UX than normal pager. This was considered as a regression compared to v.1

### Test plan

Run the example and then `CustomIndicatorExample`


src/IOSPager.tsx
```diff
@@ -0,0 +1,175 @@
...
+    if (this.props.layout.width) {
+      this.scrollTo(
+        this.props.navigationState.index * this.props.layout.width,
+        false
+      );
+    }
+  }
+
+  componentDidUpdate(prevProps: Props<T>) {
+    const amount = this.props.navigationState.index * this.props.layout.width;
```
WoLewicki:
```suggestion
    const offset = this.props.navigationState.index * this.props.layout.width;
```

README.md
```diff
@@ -285,6 +285,12 @@ If this is not specified, the default tab bar is rendered. You pass this props t
 
 Callback which returns a custom React Element to use as pager:
 
+##### `enabledIOSScrollViewPager`
```
satya164: I think no need for a separate prop. Isn't `renderPager` enough for it?

README.md
```diff
@@ -285,6 +285,12 @@ If this is not specified, the default tab bar is rendered. You pass this props t
...
+When `true` renders `ScrollView` instead of default pager. It might deliver slightly better experience. However, `gestureHandlerProps` have no impact then. Defaults to `false`. 
+
+##### `overscroll`
```
satya164: Same as above. Props could be passed to the pager directly

src/IOSPager.tsx
```diff
@@ -0,0 +1,181 @@
...
+  initialOffset: { x: number; y: number };
+};
+
+declare module 'react-native' {
+  export interface ScrollViewProps {
+    scrollToOverflowEnabled?: boolean;
+  }
+}
+
+export default class IOSPager<T extends Route> extends React.Component<
```
satya164: Can we call it `ScrollPager` instead?

src/IOSPager.tsx
```diff
@@ -0,0 +1,181 @@
...
+import { Route, Listener } from './types';
+import { Props } from './Pager';
+
+const { event, divide } = Animated;
+
+type State = {
+  initialOffset: { x: number; y: number };
+};
+
+declare module 'react-native' {
```
satya164: Not sure about this. Shouldn't this be fixed in RN?

src/IOSPager.tsx
```diff
@@ -0,0 +1,181 @@
...
+  }
+
+  private initialOffset = {
+    x: this.props.navigationState.index * this.props.layout.width,
+    y: 0,
+  };
+
+  private scrollViewRef: React.RefObject<
+    Animated.ScrollView
+  > = React.createRef();
```
satya164: `private scrollViewRef = React.createRef<Animated.ScrollView>();` might be enough

src/DefaultPager.tsx
```diff
@@ -0,0 +1,15 @@
+import React from 'react';
```
satya164:
```suggestion
import * as React from 'react';
```

src/IOSPager.tsx
```diff
@@ -0,0 +1,181 @@
+import * as React from 'react';
+import { StyleSheet, Keyboard } from 'react-native';
+import Animated from 'react-native-reanimated';
+
+import { Route, Listener } from './types';
```
satya164: nit: type imports at end

src/IOSPager.tsx
```diff
@@ -0,0 +1,181 @@
...
+      children,
+      layout,
+      onSwipeStart,
+      onSwipeEnd,
+      overscroll,
+      navigationState,
+    } = this.props;
+
+    return children({
+      position: divide(this.position, layout.width),
```
satya164: Maybe we need to move this node to a property to not create new node every render to avoid the reanimated bridge issue?

PR #956, opened 2020-01-28
chore: update @types/react-native and remove TODO

Naturalclar:
### Motivation

I've updated `@types/react-native` to fill in the missing props for ScrollView https://github.com/DefinitelyTyped/DefinitelyTyped/pull/41862
This can remove the TODO comment for adding missing props.

### Test plan
yarn typescript


satya164:
Thanks for the PR

PR #964, opened 2020-02-18
fix: exported Pager, PagerConsumer, PagerContext, and PagerContextType

immackay:
### Motivation

This allows child PanGestureHandlers to call addGestureHandlerRef using PagerConsumer (and PagerContext), and additionally allows them to waitFor the Pager PanGestureHandler through the addition of gestureHandlerRef to Pager.providerVal

#### Problems Solved
Using a PanGestureHandler inside a MaterialTopTabNavigator from react-navigation steals pan gesture input from the Pager used by TabView. 

### Example Usage

Inside a MaterialTopTapNavigator with `swipeEnabled = true`
```tsx
import React, { Component } from 'react';
import { PanGestureHandler } from 'react-native-gesture-handler';
import Animated from 'react-native-reanimated';
import { PagerConsumer, PagerContextType } from 'react-native-tab-view';

class Example extends Component {
  private pager: PagerContextType | undefined;
  private panGestureHandlerRef = React.createRef<PanGestureHandler>();
  private y = new Animated.Value(0);

  componentDidMount() {
    this.pager?.addGestureHandlerRef(this.panGestureHandlerRef);
  }

  private handlePan = Animated.event([{
    nativeEvent: ({ translationY: y }: { translationY: number }) =>
      Animated.set(this.y, y)
  }]);

  render() {
    return (
      <PagerConsumer>
        {(pager?: PagerContextType) => {
          this.pager = pager;
          return (
            <View style={{ flex: 1 }}>
              <PanGestureHandler
                ref={this.panGestureHandlerRef}
                simultaneousHandlers={[pager!.gestureHandlerRef]}
                waitFor={pager!.gestureHandlerRef}
                onGestureEvent={this.handlePan}
                onHandlerStateChange={this.handlePan}
              />
            </View>
          );
        }}
      </PagerConsumer>
    );
  }
}
```

### DIFF
--- a/src/Pager.tsx
+++ b/src/Pager.tsx
@@ -82,7 +82,14 @@ const {
   timing,
 } = Animated;
 
-const PagerContext = React.createContext({});
+export type PagerContextType = {
+  gestureHandlerRef: React.RefObject<PanGestureHandler>;
+  addGestureHandlerRef: (ref: React.RefObject<PanGestureHandler>) => void;
+};
+export const PagerContext = React.createContext<PagerContextType | undefined>(
+  undefined
+);
+export const PagerConsumer = PagerContext.Consumer;
 
 const TRUE = 1;
 const FALSE = 0;
@@ -233,9 +240,15 @@ export default class Pager<T extends Route> extends React.Component<
 
   static contextType = PagerContext;
 
+  // PanGestureHandler ref used for coordination with parent handlers
+  private gestureHandlerRef: React.RefObject<
+    PanGestureHandler
+  > = React.createRef();
+
   // Mechanism to add child PanGestureHandler refs in the case that this
   // Pager is a parent to child Pagers. Allows for coordination between handlers
   private providerVal = {
+    gestureHandlerRef: this.gestureHandlerRef,
     addGestureHandlerRef: (ref: React.RefObject<PanGestureHandler>) => {
       if (!this.state.childPanGestureHandlerRefs.includes(ref)) {
         this.setState((prevState: ComponentState) => ({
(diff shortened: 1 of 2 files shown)


osdnk:
Thanks!

CherryMan:
@satya164 Any reason this was reverted?

satya164:
It's not something we want to expose as public API because there are issues with it, like: not being able to remove a ref once you add it, `PagerContext` is the wrong name for gesture handler context, `PagerConsumer` doesn't need to be exported separately.

immackay:
> `PagerContext` is the wrong name for gesture handler context
This was the name that was already in place, I just exported it. Perhaps changing the name would be a better solution?

> `PagerConsumer` doesn't need to be exported separately
Sure, definitely, but the general standard is to export Consumer and Provider components for ease of use. This is still not a reason to revert completely however, you could just remove this export very easily.

Reverting this pull request breaks code that needs to use a PanGestureHandler inside another container that uses GestureHandlers. Can you explain any more specific issues with this PR?

immackay:
My alternative to using this code is to exclude this library altogether as without this, it does not work. Exporting the context and passing down the GestureHandlerRef should be expected of this library as it is necessary for any advanced usage, and I was both surprised and offended to find that it was not.

immackay:
> not being able to remove a ref once you add it

Is this a common behaviour? Adding this functionality should be approximately three lines of code, if that. Is this a valid reason to revert this PR?

osdnk:
We believe that this change breaks the OCP of the library and we're strongly motivated to not move with having these exports included in the lib. 
Can you try maybe come up with some another solution which won't move out of the internals of the library? 
Frankly speaking, I have problems with understanding your issue? Can you maybe describe a bit deeper what's your problem?

immackay:
"I have problems with understanding your issue"
> Using a PanGestureHandler inside a MaterialTopTabNavigator from react-navigation steals pan gesture input from the Pager used by TabView.

This means that adding extra gesture functionality inside a MaterialTopTabNavigator causes the left/right swiping in the TabNavigator to be broken. 

Please run the example I provided within a MaterialTopTabNavigator. With this PR, the code runs just fine, but without it breaks.

The other solution is to not use this library, which would be a shame.

PR #968, opened 2020-02-19
Allow child to be clickable

chehsing:
### Motivation
Allows touchable child be clickable, while not affecting parent clickability.


satya164:
Thanks for the PR. Can you describe why you need it? The reason it's disabled is because it affects the ripple position on Android, and it's unusual to have another touchable there instead of using the `onTabPress` prop.

satya164:
Closing due to no response.

PR #974, opened 2020-02-21
Make `jumpToIndex` public

todorone:
### Motivation

Sometimes, you may need to slide to the next/previous item from the component that is higher in hierarchy than `TabView`. By making `jumpToIndex` public, we get rid of TypeScript error when calling it by ref - `tabView.current.jumpToIndex(i + 1)`

### Test plan

Works perfectly 😉


osdnk:
Can you explicitly use the `public` keyword?

satya164:
You own the navigation state. To change the screen, you just need to change the `index` in your state. No need to use internal properties.

Issue #969, opened 2020-02-19
export { default as TabView, Props as TabViewProps } from './TabView' in JS build


thecodehunter:
# Current behaviour

WARNING in JS build:
export { default as TabView, Props as TabViewProps } from './TabView' in JS build
Also the same in TabBar, i checked the TS code and looks good to me, the prop is exported correctly so it looks like a build problem.

### Expected behaviour
No Warnings, Props should be exported on the JS build.

### Screenshots (if applicable)

![Screenshot from 2020-02-19 11-24-13](https://user-images.githubusercontent.com/2933118/74853386-dd412080-530b-11ea-841e-e83b5992498d.png)

### Your Environment

| software                     | version
| ---------------------------- | -------
| ios or android               | web
| react-native                 | 0.61.4
| react-native-tab-view        | 2.13.0
| react-native-gesture-handler | 1.5.6
| react-native-reanimated      | 1.4.0
| node                         | 10.16
| npm or yarn                  | 6.9

raajnadar:
As a temporary workaround use version `2.10.0` or `2.10.2`

patrickgodinho:
I have some problem using react-navigation-tabs. Is there any solution?

MathieuUrstein:
@patrickgodinho here is a few solutions:
- Hack your package-lock.json to use a working version as mentioned by @raajnadar ([example](https://stackoverflow.com/a/48524488/2544631))
- Patch the package using [patch-package](https://github.com/ds300/patch-package)

Example of patch:
Path: `patches/react-native-tab-view+2.13.0.patch`
```Diff
diff --git a/node_modules/react-native-tab-view/lib/module/index.js b/node_modules/react-native-tab-view/lib/module/index.js
index 629b89f..ed34cc0 100644
--- a/node_modules/react-native-tab-view/lib/module/index.js
+++ b/node_modules/react-native-tab-view/lib/module/index.js
@@ -1,2 +1,2 @@
-export{default as TabView,Props as TabViewProps}from'./TabView';export{default as TabBar,Props as TabBarProps}from'./TabBar';export{default as SceneMap}from'./SceneMap';export{default as ScrollPager}from'./ScrollPager';export{Route,NavigationState,SceneRendererProps}from'./types';
[...]

Issue #945, opened 2020-01-02
npm issue ('react-native-tab-view'), not able to import { tabBar } from react-native-tab-view


umairbinqaisar-sl:
### Code sample

For Syntax Highlighting check this [link](https://help.github.com/en/articles/creating-and-highlighting-code-blocks)

### Your Environment

| software                     | version
| ---------------------------- | -------
| ios or android               |
| react-native                 |
| react-native-tab-view        |
| react-native-gesture-handler |
| react-native-reanimated      |
| node                         |
| npm or yarn                  |

javadi69:
You should use `import { TabBar } from 'react-native-tab-view';` instead of `import { tabBar } from 'react-native-tab-view';`

umairbinqaisar-sl:
i did that too but the TabBar itself is showing errors, i actually uninstalled and installed the single module even tried to re-install the whole node modules but nothing seems working, i am using "react-native-tab-view": "^1.4.1", idk why the file keeps on showing errors of every class imported from react-native-tab-view.

borisvanschooten:
I have the same problem.  I do not use react-native-tab-view directly, it is an indirect import from react-native-tabs.  So this breaks react-native-tabs for me.  I get the error:

Failed to compile.

./node_modules/react-native-tab-view/lib/module/index.js
"export 'Props' (reexported as 'TabBarProps') was not found in './TabBar'

My guess is this has something to do with your use of .tsx files.

benjipott:
@borisvanschooten same issue on latest `2.12.0`

WazzaJB:
I arrived here because this error was being thrown in "react-navigation-tabs", it is worth noting that if you are using yarn you can create a version resolution to work around this:

```
"resolutions": {
    "react-navigation-tabs/react-native-tab-view": "2.11.0"
 },
```

Thanks for helping me get this solved.

PR #981, opened 2020-03-07
fix: type export warning in webpack

connectdotz:
### Motivation

fixed babel typescript transpile error for Props type exports (#969, #945):

```
./node_modules/react-native-tab-view/lib/module/index.js
"export 'Props' (reexported as 'TabBarProps') was not found in './TabBar'
```

This is mainly caused by a well-known babel typescript transform limit that it can not tell types from values, see babel/babel#8361

 Change summary
- added `isolatedModules:true` in tsconfig.json so typescript will prevent such problem going forward
- refactor the `index.ts` to explicitly export Props as types.

### Test plan

the `example` now worked

### DIFF
--- a/src/index.tsx
+++ b/src/index.tsx
@@ -1,10 +1,18 @@
-export { default as TabView, Props as TabViewProps } from './TabView';
-export { default as TabBar, Props as TabBarProps } from './TabBar';
-export {
-  default as TabBarIndicator,
-  Props as TabBarIndicatorProps,
-} from './TabBarIndicator';
+import * as tv from './TabView';
+import * as tb from './TabBar';
+import * as tbi from './TabBarIndicator';
+import * as types from './types';
+
+// help babel to transpile the types correctly
+export type TabViewProps<T extends types.Route> = tv.Props<T>;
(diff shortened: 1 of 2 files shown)


chayancbnits:
> ### Motivation
> fixed babel typescript transpile error for Props type exports (#969, #945):
> 
> ```
> ./node_modules/react-native-tab-view/lib/module/index.js
> "export 'Props' (reexported as 'TabBarProps') was not found in './TabBar'
> ```
> 
> This is mainly caused by a well-known babel typescript transform limit that it can not tell types from values, see [babel/babel#8361](https://github.com/babel/babel/issues/8361)
> 
> Change summary
> 
> * added `isolatedModules:true` in tsconfig.json so typescript will prevent such problem going forward
> * refactor the `index.ts` to explicitly export Props as types.
> 
> ### Test plan
> the `example` now worked

same issue I am facing
Screenshot from 2020-04-03 14-02-25
Screenshot from 2020-04-03 14-02-13
Screenshot from 2020-04-03 14-01-59

satya164:
Thanks a lot for the PR.

Issue #993, opened 2020-03-28
[iOS] ScrollPager and onIndexChange


solomonhawk:
### Current behaviour

With a simple slider like the example, adding the sample code from the readme to enable the `ScrollPager` on iOS results in `onIndexChange` never getting called. Is this intended (necessary for some reason that's not immediately obvious)?

### Expected behaviour

I expect `onIndexChange` to still get called when passing `renderPager={props => <ScrollPager {...props} />}`.

### Code sample

Here's an example snack that demonstrates the issue. **Make sure you are viewing this on the iOS device emulator**. Try uncommenting line 31 (`// renderPager={props => <ScrollPager {...props} />}`) in `App.js`.

Swiping between the tabs results in an alert with the index (`onIndexChange` fires `alert`). After uncommenting the line to enable usage of the `ScrollPager`, the alert no longer fires.

https://snack.expo.io/HkQl8ETI8

### What have you tried

I looked around to see whether there was another way I could observe the ending index after a swipe but didn't find a good way to do so. I can pass `onSwipeEnd` to `ScrollPager` but it gets called with no parameters and I'm left only with access to the `Animated.Value` `position` which it seems like I can't just synchronously read on the JS thread (am I wrong on that)?

#### Possible Solutions

1. Don't use `ScrollPager` - this would be a bummer because it does give a bit of a nicer interaction on iOS
2. ??? Taking a look at which animated node pipeline processing logic we might be able to pull in from `Pager` to observe `position` and `call` back out.

### Your Environment

| software                     | version
| ---------------------------- | -------
| ios or android               | ios 11
| react-native                 | 0.61.5
| react-native-tab-view        | 2.13
| react-native-gesture-handler | 1.6.1
| react-native-reanimated      | 1.7.1
| node                         | 12.13.0
| npm or yarn                  | yarn 1.21.1

satya164:
cc @osdnk

solomonhawk:
I'm guessing, naively, that part of this is due to the fact that when using `ScrollPager`, the x pan position is "owned" by the `ScrollView` and we synchronize an animated value with its `contentOffset.x` during `onScroll`. This should still provide an opportunity to `call` back out when `position` comes to a rest.

solomonhawk:
I poked around a bit and found something that patches the issue but I'm not familiar enough with things to know whether it is correct.

```js
<Animated.Code
  exec={onChange(
    this.position,
[...]

PR #994, opened 2020-03-28
fix: teach ScrollPager how to call onIndexChange

solomonhawk:
cc @osdnk @satya164 

### Motivation

I opened #993 after observing that `onIndexChange` does not appear to work when passing `renderPager={props => ScrollPager {...props } />`. This component only calls `onIndexChange` when it's `jumpTo` API method is called, but not simply as a result of scrolling within the inner `ScrollView`.

### Test plan

I have no idea but it doesn't look like there are any existing tests so I imagine this isn't a blocker. I would love to know how to test `ScrollView` interactions though so if anyone has insight please share.

### DIFF
--- a/src/ScrollPager.tsx
+++ b/src/ScrollPager.tsx
@@ -4,7 +4,7 @@ import Animated from 'react-native-reanimated';
 import { Props } from './Pager';
 import { Route, Listener } from './types';
 
-const { event, divide, Value } = Animated;
+const { event, divide, onChange, cond, eq, round, call, Value } = Animated;
 
 type State = {
   initialOffset: { x: number; y: number };
@@ -136,6 +136,7 @@ export default class ScrollPager<T extends Route> extends React.Component<
       onSwipeStart,
       onSwipeEnd,
       overscroll,
+      onIndexChange,
       navigationState,
     } = this.props;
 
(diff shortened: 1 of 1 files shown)


src/ScrollPager.tsx
```diff
@@ -188,6 +189,16 @@ export default class ScrollPager<T extends Route> extends React.Component<
           ref={this.scrollViewRef}
         >
           {children}
+          <Animated.Code
+            exec={onChange(
+              this.relativePosition,
+              cond(eq(round(this.relativePosition), this.relativePosition), [
```
solomonhawk: This has 0 tolerance for rounding errors or anything that might cause `this.relativePosition` to end up on a value that isn't an integer. Perhaps a better check would be to test whether `this.relativePosition` is very close to a round number. Using the logic from the following psuedo-code:

```
p = relativePosition
threshold = .005

atRest = p / round(p) <= (1 + threshold) && p / round(p) >= (1 - threshold)
```

```
and(
  lessOrEq(divide(this.relativePosition, round(this.relativePosition)), 1 + threshold),
  greaterOrEq(divide(this.relativePosition, round(this.relativePosition)), 1 - threshold)
)
``` 

/shrug, Thoughts?

solomonhawk:
This seems like it solves a bug in the current iteration of the library where following the directions in the README in order to use the `ScrollPager` renders the `TabView` non-functional since its navigationState index is never updated.

Wanted to bump this real quick and see whether anyone has time for a quick review.

Thanks for the time!

satya164:
I'm gonna merge this. Lemme know @osdnk if you have thoughts.

Evan-Younan:
The ScrollPager still does not scroll properly at the moment when a tab is pressed. I've tested it with the current package and with `@react-navigation/material-top-tabs`.
bug-gif

Current relevant dependencies: 

```
    "react": "^16.13.1",
    "react-native": "^0.63.0-rc.1",
    "react-native-tab-view": "^2.14.0",
    "@react-navigation/material-top-tabs": "^5.2.2",
```

satya164:
@Evan-Younan can you open an issue about it?

Evan-Younan:
Sure. #1011

osdnk:
It's def broken after these changes

PR #997, opened 2020-04-04
fix: fix scrollbars being always visible in TabBar on web

arnnis:
### Motivation
On windows there are useless scrollbar around TabBar (react-native-web). I don't know what is the purpose of `overflow:scroll` here but changing to `overflow:hidden` fixes the issue. on mac it is reproducible if you set _show scroll bars_ to _Always_ in _Settings -> General_ Which this is the default behaviour on windows

### Test plan
1. Create a simple tab view
2. You should see the scrollbars
 

image

### DIFF
--- a/src/TabBar.tsx
+++ b/src/TabBar.tsx
@@ -472,7 +472,7 @@ const styles = StyleSheet.create({
     flex: 1,
   },
   scroll: {
-    overflow: 'hidden',
+    overflow: Platform.select({ default: 'scroll', web: undefined }),
   },
   tabBar: {
     backgroundColor: '#2196f3',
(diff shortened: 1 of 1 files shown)


src/TabBar.tsx
```diff
@@ -472,7 +472,7 @@ const styles = StyleSheet.create({
     flex: 1,
   },
   scroll: {
-    overflow: 'scroll',
+    overflow: 'hidden',
```
satya164, line 475: Hey! Thanks for the PR, but this breaks React Native.
```suggestion
    overflow: Platform.select({ default: 'scroll', web: undefined }),
```

PR #1003, opened 2020-05-01
feat: add `renderTabBarItem` prop to `TabBar`

jtomaszewski:
### Motivation

We want to use `react-native-tab-view` and its' `TabBar` without replacing it, but want to replace `TabBarItem`, as we are required to do a few changes to the tab labels, that can't be achieved by just passing `labelStyle` to it. (For example, our tab animations will be changing other attribute than `opacity` which is currently set by native `TabBarItem`.)

This simple change lets replace the TabBarItem component if somebody needs to.

### DIFF
--- a/README.md
+++ b/README.md
@@ -527,6 +527,10 @@ Function which takes an object with the current route, focused status and color 
 />
 ```
 
+##### `renderTabBarItem`
+
+Function which takes a `TabBarItemProps` object and returns a custom React Element to be used as a tab button.
+
 ##### `renderIndicator`
 
 Function which takes an object with the current route and returns a custom React Element to be used as a tab indicator.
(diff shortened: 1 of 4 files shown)


jtomaszewski:
> Can you change it to call it like renderTabBarItem({ ... }) instead of <TabBarItem />? Since we're not accepting a component here, but a render callback, we shouldn't use it like a component. Otherwise it'll remount the component every time the reference to the callback changes.

I assumed a developer would just pass a component or a memoized callback here, and so that problem wouldn't exist ; P But on the other hand it's easy to transform a component into a render prop like this:

```tsx
const renderTabBarItem = (props): React.ReactElement => (
  <AiloTabBarItem {...props} />
);
```

So sure, I can do that change :) I force pushed that into this branch. I also added minor docs about that new option to README.

satya164:
Merged in 33fd8087a03299edeac7d7c79701e2622423f08a

PR #1010, opened 2020-05-12
fix: use `currentlyFocusedInput` for latest RN

MaxEgg:
Replaced the currentlyFocusedField method with currentlyFocusedInput method. This fixes the error "currentlyFocusedField is deprecated and will be removed in a future release. Use currentlyFocusedInput".

### Motivation
CurrentlyFocusedField is deprecated.
Error being thrown when swiping with react-navigation.


src/Pager.tsx
```diff
@@ -621,7 +621,7 @@ export default class Pager<T extends Route> extends React.Component<
               this.interactionHandle = InteractionManager.createInteractionHandle();
 
               if (keyboardDismissMode === 'auto') {
-                const input = TextInput.State.currentlyFocusedField();
+                const input = TextInput.State.currentlyFocusedInput();
```
satya164, line 624:
```suggestion
                const input = TextInput.State.currentlyFocusedInput
                	? TextInput.State.currentlyFocusedInput()
                	: TextInput.State.currentlyFocusedField();
```

satya164:
Thanks

PR #1075, opened 2020-09-16
fix: fix 'Attempted import error' in 'react-native-reanimated'

indiev:
image

./node_modules/react-native-tab-view/lib/module/TabBarIndicator.js
Attempted import error: 'EasingNode' is not exported from 'react-native-reanimated'.

an error occurred in react-native-web + nextjs


jackphuongvu:
Any updates here?

Issue #441, opened 2018-01-31
White space when scrolling inside a scene and swiping to the next one (with collapsible Header)


Crash--:
### Expected behaviour

If I scroll inside a scene, I expect that after swiping to an another one, the content is well displayed. With no white / blank space. 

### Actual behaviour

Use this project  : https://snack.expo.io/@satya164/collapsible-header-with-tabview 

Scroll inside the first scene then swipe right. See this space ? 

![ezgif-1-e5e37a27dd](https://user-images.githubusercontent.com/1107936/35625773-cbf22d72-0693-11e8-99c5-1f8f301ad299.gif)

### What have you tried
scrollTo with/without timeout() with no success.

austenLacy:
@Crash-- I had a similar issue and solved it using the `onScroll` event listener

```javascript
```

sircharleswatson:
@Crash-- Any luck with this yet? I'm facing the same issue at the moment. not sure how to resolve it.

sircharleswatson:
@Crash-- I figured out a solution for this but it's not dynamic (have to set it up for each tab individually) I'm gunna try to do a repro on Expo soon but I'm super busy at the moment. I'll keep this thread up-to-date when I do it!

fernandoantunes:
@sircharleswatson What solution did you find?

harrie006:
@sircharleswatson What was the solution?

sircharleswatson:
The solution I came up with was super convoluted and ridiculous (even though it worked) but I had to redesign away from using tabs on this particular page because it was causing too many other issues. In conclusion, I will try to avoid animated headers + tabs as much as possible in the future 🙈

austenLacy:
@fernandoantunes @harrie006 @sircharleswatson I have a working (with some kinks) example that fixes this problem -- https://github.com/austenLacy/react-native-collapsible-header-with-tabs-demo

satya164:
Hey, I just released a new alpha `2.0.0-alpha.0` of the library. It's rewritten using [react-native-gesture-handler](https://github.com/kmagiera/react-native-gesture-handler) and [react-native-reanimated](https://github.com/kmagiera/react-native-reanimated) addresses a many platform specific bugs and performance problems. The documentation is updated as well.

Please try the new version and see if it addresses your issue. If not, please open a new issue following the issue template.

youngsaksham:
[...]

Issue #411, opened 2017-12-19
【help】how can i resolve sticky tab view inside scrollview


YouHan26:
Found a same issue, but not I want.

### Expected behaviour
I have three flatlist using react-native-tab-view. And the tab view inside parent scrollview.
when the tab view is not scroll to header of screen, the parent scrollview will always scroll even if touch the tab view content.
But after the tab view scroll to header of screen, the child flatlist of tab view can scroll auto.
When drag down, the child flatlist will scroll first. And when child flatlist scroll to top, parent scrollview will start to scroll down. 

### Actual behaviour
none
### Code sample, screenshots
none
### What have you tried
I use `scrollEnabled` attribute to resolve this case. When tab view scroll to header of screen. Set parent scrollendabled false and child flatlist scrollendable true.
It does work. But it always pause when tab view scroll to header or leave.
Because only one component can response panresponder event at the same time.
Any advice ? Thanks very much and pardon my poor English...

onlybenyang:
Maybe the stickyHeaderIndices of ScrollView would give u a  hand. However, you need render the tab header yourself in <ScrollView>

Crash--:
We arrive at the the same conclusion: i.e. dealing ourself with the tab header. 

I was expecting that using `<TabBar>` outside of `<TabViewAnimated>` would be trivial. But no. We can't use `<TabBar>` and we need to rewrite all this stuff. It's pretty bad since the `<TabBar>` component deal with a lot of use case... 

Is there a way to make  `<TabBar>` outside a `<TabViewAnimated>` ?

satya164:
Don't nest the TabView inside a ScrollView. The FlatList cannot optimize it's rendering when you do that. Listen to scroll events and apply translation to the tab bar to make it sticky.

Example: https://snack.expo.io/@satya164/collapsible-header-with-tabview

happyjosh:
@satya164 How to set same offset when there are multiple page?

ShabareeshSunku:
@satya164 in the expo snack you mentioned above, once the header is collapsed and if we navigate to another tab, can see some whitespace on the tab, coz of padding being set. Any workaround to fix that issue?

YouHan26:
@satya164 now I came a nested tab view case. I have a animated scroll tab view like below above, and one of its tab have another tab view. The child tab view will sticky follow parent tab,  I want to use scroll y to manual control scrollenabled, but it's very trouble, have any advice ? thanks sooo much~

[...]

PR #1079, opened 2020-09-26
Collapsible Header With Tab View

jehartzog:
### Motivation

The UI of having a collapsible header above a Tab View is a commonly requested pattern (#411), but it is surprisingly challenging for newer RN devs to implement. Creating a wrapper for this functionality and some examples for using the wrapper will help allow them to have a baseline for how to add this pattern into their own apps.

#### Prior Examples

@satya164 Provided https://snack.expo.io/@satya164/collapsible-header-with-tabview as an example. It's a valuable start for understanding how to work on this, but the snack itself can't be used without significant additional work due to the additional of large amounts of padding above the content when switching tabs in certain cases #441.

[This example](https://blog.expo.io/implementation-complex-animation-in-react-native-by-example-search-bar-with-tab-view-and-collapsing-68bb43be2dcb) appears to be complete, but it's a rather larger example that is challenging to quickly bring into a different projects that just wants the minimum code needed to reproduce the baseline effect.

@JungHsuan Provided a [medium article](https://medium.com/@linjunghsuan/implementing-a-collapsible-header-with-react-native-tab-view-24f15a685e07) on his approach, which was practically production ready and had the exact UI flow that was buttery smooth and consistent in all cases. I started with this approach, and converted it into something more re-usable in different situations, trying to make it quicker for other developers to pull into their projects.

### UI Animation

Sep-26-2020 13-36-27

### Current known limitations/issues

1. Since the collapsible header view becomes absolutely positioned and is animated, you'll need to adjust `zIndex` of the headers above it (see example code)

1. In order to support scrolling the `TabBar` up, we must set a minimum height for the scrollable content. If you have one tab with only a small amount of content in there, it will be allowed to scroll up with large padding below it. This maybe be possible fix via measuring the content, but that would be an enhancement. Edit: Discussed possible fix [here](https://github.com/JungHsuan/react-native-collapsible-tabview/issues/9#issuecomment-699583993).

1. The `CollapsibleHeader` `prop` structure I initially proposed may not be ideal. It basically forces use to create an array of `renderItem` and `data`, which is used by each tab separately to draw its content. But the idea behind this even for non-virtualized content, we need a `ScrollView` or `FlatList` wrapper in there to drive this behavior, so we want to bake this functionality into the wrapper while still making implementation lightweight. This also has the side effect of fully supporting both tab content that are rendered at once, or using `FlatList` virtualized rendering.

1. I attempted to use `SectionList` to drive this, but challenges with measuring scroll distances shut down my early attempts.

1. If you scroll tab 1 all the way down, swap to tab 2, scroll it up any amount, and come back to tab 1, you'll notice tab 1 is now at the top, with the collapsible header being set at the same height. This isn't a real 'limitation' as it's simply necessary for the UI to be smooth, but it's good to be aware of.

### Test plan

This is an early draft version of the PR. I am submitting in this state to see if there is interest in me polishing it up.

TODO's:
- [ ] Stricter typings (I used `any` in many places where it can be better typed)
- [ ] Removal of inline functions
- [ ] Adding example of correctly using underlying `FlatList`
- [ ] There is some warning spam right now that needs to be cleaned up

### DIFF
--- a/example/src/App.tsx
+++ b/example/src/App.tsx
@@ -23,6 +23,7 @@ import TabBarIconExample from './TabBarIconExample';
 import CustomIndicatorExample from './CustomIndicatorExample';
 import CustomTabBarExample from './CustomTabBarExample';
 import CoverflowExample from './CoverflowExample';
+import CollapsibleHeaderExample from './CollapsibleHeaderExample';
 
 type State = {
   title: string;
@@ -51,6 +52,7 @@ const EXAMPLE_COMPONENTS: ExampleComponentType[] = [
   CustomIndicatorExample,
   CustomTabBarExample,
   CoverflowExample,
+  CollapsibleHeaderExample,
 ];
 
 const KeepAwake = () => {
@@ -235,6 +237,7 @@ const styles = StyleSheet.create({
   },
   appbar: {
     borderBottomColor: 'rgba(0, 0, 0, 0.1)',
+    zIndex: 1, // Needed for keeping the screen header above the collapsible tab view header
   },
   content: {
     flexDirection: 'row',
--- a/example/src/CollapsibleHeaderExample.tsx
+++ b/example/src/CollapsibleHeaderExample.tsx
@@ -1,0 +1,121 @@
+import * as React from 'react';
+import { StyleSheet, View, Text } from 'react-native';
+import {
+  TabBar,
+  NavigationState,
+  SceneRendererProps,
+} from 'react-native-tab-view';
+import Article from './Shared/Article';
+import Albums from './Shared/Albums';
+import Contacts from './Shared/Contacts';
+
+import CollapsibleHeader from '../../src/CollapsibleHeader';
+
+type State = NavigationState<{
+  key: string;
+  title: string;
+}>;
+
+// These must be hardcoded to support the various animations needed for the effect
+const HEADER_HEIGHT = 144;
+const TAB_BAR_HEIGHT = 48;
+
+export default class CollapsibleHeaderExample extends React.Component<
+  {},
+  State
+> {
+  // eslint-disable-next-line react/sort-comp
+  static title = 'Collapsible header tab bar';
+  static backgroundColor = '#3f51b5';
+  static appbarElevation = 0;
+
+  state = {
+    index: 1,
+    routes: [
+      { key: 'article', title: 'Article' },
+      { key: 'contacts', title: 'Contacts' },
+      { key: 'albums', title: 'Albums' },
+    ],
+  };
+
+  private handleIndexChange = (index: number) =>
+    this.setState({
+      index,
+    });
+
+  private renderTabBar = (
+    props: SceneRendererProps & { navigationState: State }
+  ) => (
+    <TabBar
+      {...props}
+      scrollEnabled
+      indicatorStyle={styles.indicator}
+      style={styles.tabbar}
+      tabStyle={styles.tab}
+      labelStyle={styles.label}
+    />
+  );
+
+  render() {
+    return (
+      <CollapsibleHeader
+        // This is dummy data, we render the entire content as a single component for both tabs
+        // TODO: Create an example which explains why we do it this way, by showing proper use of the underlying FlatList
+        tabData={[[{ key: '0' }], [{ key: '1' }], [{ key: '2' }]]}
+        tabBarHeight={TAB_BAR_HEIGHT}
+        headerHeight={HEADER_HEIGHT}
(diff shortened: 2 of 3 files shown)


sagrawalcb (approved):

sagrawalcb (approved):

hejun041:
When the header's height > screen height it cant work
because the header cant scroll

zyslife:
You can try the [react-native-head-tab-view.](https://github.com/zyslife/react-native-head-tab-view)
(tabs with collapsible header)

zyslife:
@jehartzog @hejun041 @sagrawalcb @sjmueller @fantattitude 
You can try the [react-native-head-tab-view.](https://github.com/zyslife/react-native-head-tab-view)
(tabs with collapsible header)

jehartzog:
@zyslife Thanks for putting that together. In this case, I'm hoping to add the functionality as an optional inclusion to `react-native-tab-view` as I think that will help the larger base of devs who already use `react-native-tab-view` in their RN project and what to support this behavior.

Is there any interest in merging this in? I'm still happy to clean this up some more, write docs, but I've learned to wait until I hear back from maintainers.

alexco2:
I´m trying to implement this solution with SectionList right now. Regarding the limitations of point 4, are you referring to the fact that the only scrollTo function that is supportted in sectionList is the scrollToLocation one?

jehartzog:
@alexco2 Yes, that is what I saw. I only looked for ~10 minutes into it, as I suspected the `scrollToLocation` would be more complicated due to being dependent on what kind of data was added. I didn't really look into it deeply though.

alexco2:
@jehartzog `scrollToLocation` expects `itemIndex` and `sectionIndex`. Both can be set to 0 though. With `viewOffset` you can then control the scrolling position with the use of pixel like in your example with the flatlist. Two problems then accrued. 

First, if the sectionlist of the inactive route doesn´t have enough rows so that the header can be completely collapsed, the function `syncScrollOffset` somehow updates the header of the active route with the amount of `translateY` the inactive route can be scrolled. So, if the active route was scrolled with 50 pixel but the inactive route can just be scrolled 5 pixel because of to little rows, the active collapsible header gets updated with a `translateY` value of 5. My workaround is to dynamically extend a listfooter so that the list can alwas be scrolled to collapse the header completly.

Also, if the inactive route does not have any data in it, the `syncScrollOffset` function throws an out of bounds error because the `itemIndex = 0` and `sectionIndex = 0` do not exist. Weirdly using -1 each does not fix this. My workaround is to just deactivate the second route if there is no data, since it does not have any information for the user anyways.

zyslife:
> @zyslife Thanks for putting that together. In this case, I'm hoping to add the functionality as an optional inclusion to `react-native-tab-view` as I think that will help the larger base of devs who already use `react-native-tab-view` in their RN project and what to support this behavior.
> 
> Is there any interest in merging this in? I'm still happy to clean this up some more, write docs, but I've learned to wait until I hear back from maintainers.

Thanks for your invitation,If I have enough energy, I should be able to `react-native-tab-view` add this feature. But not right now. I can only make `react-native- head-tab-View ` work right now ,out of my working hours. I hope I can do something in the future.

PedroBern:
Hi folks I'm coming late to the conversation but was also working on this, I've just made a PR, let me know if you guys see how to improve it :)

jehartzog:
We've been using the code based on this PR successfully in our prod app for a while, but going forward I'd probably pickup the outside packages published by #1096 rather than continue to maintain our fork. Just throwing this in for anyone looking to add these features.

On the other hand, repo owners decide they want to pull in this PR, I will be happy to spend the time to clean it up, get it back up to date, and get it merged. Just need someone w/ review authority to give thumbs up or down.

MonsterAnan:
@jehartzog thanks for you solve my issue！！

KrisLau:
> On the other hand, repo owners decide they want to pull in this PR, I will be happy to spend the time to clean it up, get it back up to date, and get it merged. Just need someone w/ review authority to give thumbs up or down.

@satya164

okwasniewski:
Hey @jehartzog, thanks for your PR. Can you rebase this with the main branch? So we can get back to working on this and hopefully merge it. I will review this once you rebase it.

okwasniewski:
After consulting this with @satya164 we decided that we want to keep the library as small as possible (and easy to maintain) so I suggest you to publish this as a separate library.

KrisLau:
@okwasniewski Would this feature be a huge increase in size to the library? This feature would be extremely useful but i definitely understand the reasoning haha

PR #1081, opened 2020-09-29
fix: Only update state if mounted.

milesj:
### Motivation

This component always logs error in the console when testing.

```
Warning: Can't call setState on a component that is not yet mounted. This is a no-op, but it might indicate a bug in your application. Instead, assign to `this.state` directly or define a `state = {};` class property with the desired state in the Pager component.

Warning: Can't call forceUpdate on a component that is not yet mounted. This is a no-op, but it might indicate a bug in your application. Instead, assign to `this.state` directly or define a `state = {};` class property with the desired state in the Pager component.
```

PR adds some mounted checks.

### Test plan

As long as tests pass and dont log.

### DIFF
--- a/src/Pager.tsx
+++ b/src/Pager.tsx
@@ -133,6 +133,8 @@ export default class Pager<T extends Route> extends React.Component<
   };
 
   componentDidMount() {
+    this.mounted = true;
+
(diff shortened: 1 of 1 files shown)


satya164 (approved):
Thanks!

PR #1083, opened 2020-10-02
fix: fix `TabBarIndicator` width props type

kimak:
Following the implementation https://github.com/satya164/react-native-tab-view/blob/master/src/TabBarIndicator.tsx#L123
width can be a string or a number.


satya164 (approved):
Thanks!

PR #1088, opened 2020-10-15
fix: type definition build and remove unused eslint disables

kuasha420:
'Local installation' of this package  (aka, from git url) was failing on build state because of missing dependencies. I've added those as devDependency to resolve said type definition build errors.

I've also removed unused eslint disables from the webpack.config.js of example folder which was causing precommit hook to fail. 

### Motivation

To be able to install and test latest changes from the git repo, a successful build is needed, specially when consuming the package on a TS project.

### Test plan

1. Package builds without error. 
2. Commit without precommit hook failing 

Best regards.

### DIFF
--- a/example/webpack.config.js
+++ b/example/webpack.config.js
@@ -1,7 +1,5 @@
 const path = require('path');
-// eslint-disable-next-line import/no-extraneous-dependencies
 const createExpoWebpackConfigAsync = require('@expo/webpack-config');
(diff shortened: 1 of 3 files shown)


satya164:
Thanks for the PR, but the proper way to address this is to use a separate config for build which I added in https://github.com/satya164/react-native-tab-view/commit/c9b8c22b188c06416cd453f4de0bb08760b652b2

PR #1090, opened 2020-10-29
fix: avoid loading next tab on tab mount (ios)

AmauryLiet:
### Motivation

I noticed that my 2nd tab was preloading (although in lazy=true + lazyPreloadDistance=0) on iOS on component mount
With further investigation, I noticed that the listener of 'enter' was called with a value wrongly rounded: `ceil(2.000000) = 3`

### Test plan

Simply tested again on iOS simulator, navigating between 2/3 tabs

I did **not** try to change props to explore other use cases (disabling lazy, or changing lazyPreloadDistance)

### DIFF
--- a/src/Pager.tsx
+++ b/src/Pager.tsx
@@ -601,11 +601,11 @@ export default class Pager<T extends Route> extends React.Component<
           ? lessThan(this.gestureX, 0)
           : greaterThan(this.gestureX, 0),
         // Based on the direction of the gesture, determine if we're entering the previous or next screen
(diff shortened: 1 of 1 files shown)


AmauryLiet:
This PR is worth a discussion, as the underlying issue is in react-native-reanimated floor/ceil implementation.
I reproduced the bug with their 1.8.0 version, and they changed their implementation in their 1.10.0
Worth giving it a try

Feel free to decline the PR if I don't give any news soon :)

PR #1095, opened 2020-11-09
feature: Adding option to override accessibility props

felire:
### Motivation

With this pr, now any person using this library will be able to override some of the default props of accessibility. The idea is to solve the problem of the default role that the library has right now. For example in my case, I want to remove the role and the state and using it on a different way because it doesn't behaves as I want.

The idea is not override the `accessibilityLabel` or `accessible` props, because there is a way of handling them, so I just move them on a way to prevent this. No one won't be able to override `accessibilityLabel` or `accessible` but the user will be able to override any other prop related to accessibility

### Test plan

I think that just using the library in a normal way and sending some accessibility props to modify the configuration should be enough.

### DIFF
--- a/src/TabBar.tsx
+++ b/src/TabBar.tsx
@@ -8,6 +8,7 @@ import {
   LayoutChangeEvent,
   I18nManager,
   Platform,
+  AccessibilityProps,
 } from 'react-native';
 import Animated from 'react-native-reanimated';
 import TabBarItem, { Props as TabBarItemProps } from './TabBarItem';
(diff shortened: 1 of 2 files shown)


satya164 (changes requested):
Thanks for the PR. Can you describe the issues you are having with the current accessibility related props?

It's better to understand the issue and try to fix that instead of exposing all accessibility-related props - and even if this says "accessibility props", nothing prevents the user from specifying other props and potentially breaking the library.

okwasniewski:
Closing this as stale

PR #1102, opened 2020-12-02
fix: TabBar with scrollEnable to take into consideration the content margins

andreibarabas:
If you want to use a custom TabBar like

```
<TabBar
      {...tabBarProps}
      scrollEnabled={true}
      contentContainerStyle={{marginLeft: 14}}
/>
```

currently, the last tab will be cut with 14px, as it does not take into consideration these kind of spacing.

This PR fixes that.

### Motivation

The last tab item is cut with with the amount you put as padding / margin to the contentContainer.

### Test plan

Let me know if you need any screenshots, but should be straitforward

### DIFF
--- a/src/TabBar.tsx
+++ b/src/TabBar.tsx
@@ -148,6 +148,33 @@ export default class TabBar<T extends Route> extends React.Component<
     return tabStyle ? tabStyle.width : undefined;
   };
 
+  //
+  // when calculating the scroll distance, take into consideration also any spacing from the container 
+  //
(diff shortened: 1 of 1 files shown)


satya164 (changes requested):
Thanks for the PR. I think this also needs to consider `marginHorizontal` and `paddingHorizontal`. Can you add it?

Also I'm not sure if specifying margin inside `contnentContainerStyle` would work properly. Can you verify that?

Please also add a video showing how it works.

okwasniewski:
Closing this as stale

PR #1105, opened 2020-12-10
fix: moved color to last prop

adamsythe:
### Motivation

For tabbar labels I want to add styles and either active or inactive color.  The assumption is that the color prop would override the style prop.


satya164 (approved):
Thanks

kamui545:
Not sure if it was really the "recommended" way to use it, but this will break for people using `<TabBar labelStyle={{ color: 'black' }} />`.

You now have to set both `activeColor` and `inactiveColor`, which I don't mind at all, just letting know people.

PR #1116, opened 2021-01-14
Fix delay after tabPress on Android

darron1217:
### Motivation

Fix issue #1114

### Test plan

I cannot imagine the side-effect of this merge. Please check the code.

### DIFF
--- a/src/Pager.tsx
+++ b/src/Pager.tsx
@@ -419,16 +419,17 @@ export default class Pager<T extends Route> extends React.Component<
     // A tab switch might occur when we're in the middle of a transition
     // In that case, the index might be same as before
     // So we conditionally make the pager to update the position
-    if (navigationState.index === index) {
-      this.jumpToIndex(index);
-    } else {
-      onIndexChange(index);
+    this.jumpToIndex(index);
+    if (navigationState.index !== index) {
+      requestAnimationFrame(() => {
+        onIndexChange(index);
 
-      // When the index changes, the focused input will no longer be in current tab
-      // So we should dismiss the keyboard
-      if (keyboardDismissMode === 'auto') {
-        Keyboard.dismiss();
-      }
+        // When the index changes, the focused input will no longer be in current tab
+        // So we should dismiss the keyboard
+        if (keyboardDismissMode === 'auto') {
+          Keyboard.dismiss();
+        }
+      });
     }
   };
 
--- a/src/Pager.tsx
+++ b/src/Pager.tsx
@@ -410,7 +410,7 @@ export default class Pager<T extends Route> extends React.Component<
   };
 
   private jumpTo = (key: string) => {
-    const { navigationState, keyboardDismissMode, onIndexChange } = this.props;
+    const { navigationState, keyboardDismissMode } = this.props;
 
     const index = navigationState.routes.findIndex(
       (route) => route.key === key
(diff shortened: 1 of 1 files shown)


alekaizer:
@satya164 When you get time kindly take a look at this PR

satya164 (changes requested):
Thanks for the PR, but this seems to basically skip state update on tab change making the navigation state and animation state separate from each other. The library isn't designed around being used as an uncontrolled component, which makes this problematic and prone to cause other issues as well.

If re-rendering takes time, the proper way to fix is to find what's the bottleneck and address that instead of skipping re-render.

darron1217:
@satya164  No this PR doesn't skip state update.
`jumpToIndex` calls `onIndexChange`, the only difference is this PR calls method asynchronously.

satya164:
> the only difference is this PR calls method asynchronously

So the question is why is doing it synchronously slow? The only thing re-rendering here should be the tab bar which shouldn't be taking this much time, or it needs some optimization. Seems like there's another issue that needs to be fixed.

Even if the issue doesn't happen in this particular case, navigating to another tab from a button in your own code would still exhibit this behavior.

This might just be happening due to Reanimated, which I plan to migrate away soon and use ViewPager instead.

satya164:
Hey @darron1217, I have finished migration to viewpager, can you confirm if the PR is still necessary?

nbermudezs:
Hi there! I was also experiencing some performance issues on Android when pressing on the tabs (swiping was smooth though). 

After some testing I found that performance improves when I replace the `onIndexChange` handler with something that does not change the state (e.g. `onIndexChange={console.log}`). I've been trying to understand why so I ended up looking at the source code of the migration to pager and noticed that 1) the `index` coming from `navigationState` is only used as a initial value as seen [here](https://github.com/satya164/react-native-tab-view/blob/main/src/PagerViewAdapter.tsx#L127), 2) the `onIndexChange` now runs _after_ the index has changed as seen [here](https://github.com/satya164/react-native-tab-view/blob/main/src/PagerViewAdapter.tsx#L145). 

Looking at those changes in `PagerViewAdapter` to me it seems that `TabView` is no longer a controlled component [as stated in the readme](https://github.com/satya164/react-native-tab-view/blob/main/README.md?plain=1#L157). Is this now the case? If so, would it make sense to make both `index` and `onIndexChange` at least optional? (I don't know enough to suggest removing them altogether 😅 )

gerywahyunugraha:
Hi @nbermudezs thanks a lot for your hint, changing `onIndexChange` does indeed cause the onPress to run much smoother. I'm a bit curious because setting this to the code below doesn't have any impact on the rest of the functionality.

```js
      <TabView
        lazy
        navigationState={{ index, routes }}
        renderScene={renderScene}
        renderTabBar={renderTabBar}
        onIndexChange={() => console.log('hello')}
        initialLayout={{ width: layout.width }}
      />
```

Did you settle with this solution for the problem? Or did you find another way to solve it?

nbermudezs:
Hey @gerywahyunugraha. I ended up doing 
```
onIndexChange={Platform.OS === 'web' ? setIndex : () => {}}
```
but in all honesty I don't remember if I really needed the `setIndex` for web here or if I was trying to fix something else. 

Like I mentioned in my post I think the `TabView` is no longer a controlled component so doing this should be fine and it should really be a matter of updating the docs and maybe making `onIndexChange` not required. My app has been live for a few months now and we haven't had any complaints so I don't think there are any practical problems in doing this either.

PR #1122, opened 2021-02-02
fix: interaction handle not covering the whole tab transition animation

stanislav-velichkov:
### Motivation

Currently the interaction handle covers only swipe gesture. It should cover the whole tab transition animation, so work can be delayed until it is finished and thus prevent stutters. Might be a breaking change for someone depending on the current behaviour.

### Test plan

Tested on iOS and Android, the handle is created/cleared as expected both on swipe gesture and tab press.


satya164:
Thanks for the PR and sorry for the delay. I need to close the PR since we no longer use Reanimated. Thanks again.

PR #1153, opened 2021-03-19
fix: make props with default value optional

robrechtme:
### Motivation

When using the TabView component, I get TS errors for props that I want to leave default.


robrechtme:
I see this is fixed already.

PR #1186, opened 2021-05-27
feat: add rn compatibility disclaimer table to readme.md

quixote15:
Add React Native compatibility disclaimer to readme

### Motivation

App crash using version 3.x.x of this library as in this issue https://github.com/satya164/react-native-tab-view/issues/1143

### Test plan

no need.

### DIFF
--- a/README.md
+++ b/README.md
@@ -26,10 +26,10 @@ A cross-platform Tab View component for React Native. Implemented using [`react-
 ## React Native Compatibility
 To use this library you need to ensure you are using the correct version of React Native. If you are using a version of React Native that is lower than `0.63` you will need to upgrade that before attempting to use this library.
 
-| `react-native-tab-view` version | Required React Native Version|
-| ---------| ---------|
-| `2.x.x`  | `< 0.63` |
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -22,6 +22,15 @@ A cross-platform Tab View component for React Native. Implemented using [`react-
 
 <a href="https://raw.githubusercontent.com/satya164/react-native-tab-view/main/demo/demo.mp4"><img src="https://raw.githubusercontent.com/satya164/react-native-tab-view/main/demo/demo.gif" width="360"></a>
 
+
+## React Native Compatibility
+To use this library you need to ensure you are using the correct version of React Native. If you are using a version of React Native that is lower than `0.63` you will need to upgrade that before attempting to use this library.
```
satya164, line 27:
```suggestion

To use this library you need to ensure you are using the correct version of React Native. If you are using a version of React Native that is lower than `0.63` you will need to upgrade that before attempting to use this library.
```

satya164:
Thanks for the PR!

satya164 (approved):

PR #1188, opened 2021-06-01
feat: macOS and Windows support

dcangulo:
Using `react-native-pager-view` on macOS and Windows causes errors. Solution is to use `PanResponder` for these platforms (same with web).

Those platforms favors `.native.tsx` file over `.tsx` file which caused the errors. Solution is to use platform specific components for each.

macOS fixes:
Transform property is currently broken on `react-native-macos`. Temporary solution is to use `left` for indicator position. See: https://github.com/microsoft/react-native-macos/issues/280

Windows fixes:
`PlatformPressable` is not pressable on Windows (only the label), I'm not sure why, but adding `backgroundColor: 'transparent'` to it makes the whole tab pressable.

### Motivation
`react-native-pager-view` does not support macOS and Windows.

### Test plan
Run on `react-native-macos` and `react-native-windows` projects.

Windows (initial):
windows

Windows (swiping):
windows2

macOS (initial):
Screen Shot 2021-06-01 at 11 18 18 AM

macOS (swiping):
Screen Shot 2021-06-01 at 11 19 32 AM

macOS (swiping): if  https://github.com/microsoft/react-native-macos/pull/460 is merged
Screen Shot 2021-06-01 at 12 56 17 PM

macOS (after swipe/swipe release):
Screen Shot 2021-06-01 at 11 20 11 AM


raaannzzz:
+1

dcangulo:
@satya164 Any more changes I need to make? :crossed_flags:

satya164:
I'll review it this week

satya164:
Thanks for the PR 🙏

PR #1191, opened 2021-06-10
Hide invisible tabs content (by setting their height to 0)

### LINKED ISSUE #290: Height in different scene
**How** to fix differences in height from one tab to another?

SiSa68:
There is empty space on bottom of tab's scene, if other tabs height is larger than current tab

### Motivation

Fix #290 
In this PR I hide other tabs content (by setting their contents height to 0) to resolve this problem

### Test plan

- Add 3 tabs with different content height
- Switch between tabs
- Each tabs height is same as its contents height

### DIFF
--- a/src/SceneView.tsx
+++ b/src/SceneView.tsx
@@ -15,6 +15,7 @@ type Props<T extends Route> = SceneRendererProps &
     index: number;
     children: (props: { loading: boolean }) => React.ReactNode;
     style?: StyleProp<ViewStyle>;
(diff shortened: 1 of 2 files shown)


satya164 (changes requested):
Pretty sure it won't be smooth to change the height of content when swiping, and this may cause issues with components such as FlastList.

Please provide videos showing that it works properly in all cases.

okwasniewski:
Closing this as stale

PR #1192, opened 2021-06-14
fix: RTL support for the web

volfpe:
The gesture handling on the web is broken for RTL languages. This pull request fixes the responders for RTL support.

### Motivation

`react-native-pager-view` support for RTL languages is not ideal on the web.

### Test plan

- Have at least two tabs.
- Use the web version with RTL locale.
- The gesture handling is working correctly.


satya164 (approved):
Thanks for the PR

Issue #1178, opened 2021-05-07
[v3] TabView's height does not expand anymore to the content size


TPXP:
### Current behaviour
I'm wrapping a Top Bar navigator in a scrollview to show additional content below some content. I want the user to be able to scroll down the page to see the content in the top bar navigator. Yet, in order to be able to scroll back up, I wrap all the page content (including the navigator) in a ScrollView. When using v2, the TabNavigator would expand to its content size, thus making it visible without having to implement a scrollview in the screens of the navigator. This does not seem to be the case anymore, as content is shrinked to the height of the navigator.

### Expected behaviour
I expect the navigator's height to be maximum of the screens' heights so that the content can be scrolled

### Code sample
```jsx
import React from 'react';
import {View, StyleSheet, ScrollView} from 'react-native';
import {createMaterialTopTabNavigator} from "@react-navigation/material-top-tabs";

export default function Posts() {
  const Tab = createMaterialTopTabNavigator();
  return (
    <ScrollView style={{flex:1}} contentContainerStyle={{minHeight:'100%', backgroundColor:'white'}}>
      <View style={{height:300, width:200, backgroundColor:'red'}} />
      <Tab.Navigator>
        <Tab.Screen
          name={"A"}>
          {() => <>
            <View style={{height:300, width:200, backgroundColor:'orange'}} />
            <View style={{height:300, width:200, backgroundColor:'red'}} />
            <View style={{height:300, width:200, backgroundColor:'blue'}} />
            <View style={{height:300, width:200, backgroundColor:'green'}} />
          </>}
        </Tab.Screen>

      </Tab.Navigator>
    </ScrollView>
  );
}
```

### Screenshots (if applicable)
Trying to scroll down in both cases

| v2 | v3 |
|---|---|
| ![Simulator Screen Shot - iPhone 12 - 2021-05-07 at 12 19 22](https://user-images.githubusercontent.com/7191841/117435859-9b12f780-af2e-11eb-983c-97c6f4e212b3.png) | ![Simulator Screen Shot - iPhone 12 - 2021-05-07 at 12 19 42](https://user-images.githubusercontent.com/7191841/117435887-a0704200-af2e-11eb-9a0a-aeaf6bdecd4b.png) |

### What have you tried

Downgrading to the v2 series fixes the issue
```
yarn add react-native-tab-view@2
```
... But I like my dependencies to be up to date 😄 

### Your Environment

| software                     | version
| ---------------------------- | -------
[...]

PR #1198, opened 2021-06-27
fix: Tab view height auto expand

wael-zuaiter:
### Motivation

TabView's height does not expand anymore to the content size in version 3

A different approach would be to refactor `SceneView` component to be a functional component

Resolves: #1178

### DIFF
--- a/src/SceneView.tsx
+++ b/src/SceneView.tsx
@@ -14,7 +14,7 @@ type Props<T extends Route> = SceneRendererProps &
     lazyPreloadDistance: number;
     index: number;
     children: (props: { loading: boolean }) => React.ReactNode;
-    style?: StyleProp<ViewStyle>;
+    containerStyle?: StyleProp<ViewStyle>;
   };
 
 type State = {
@@ -92,7 +92,7 @@ export default class SceneView<T extends Route> extends React.Component<
   };
 
   render() {
-    const { navigationState, index, layout, style } = this.props;
+    const { navigationState, index, layout, containerStyle } = this.props;
     const { loading } = this.state;
 
     const focused = navigationState.index === index;
@@ -110,7 +110,7 @@ export default class SceneView<T extends Route> extends React.Component<
             : focused
             ? StyleSheet.absoluteFill
             : null,
-          style,
+          containerStyle,
         ]}
       >
         {
--- a/src/TabView.tsx
+++ b/src/TabView.tsx
@@ -112,7 +112,7 @@ export default function TabView<T extends Route>({
                       lazy={typeof lazy === 'function' ? lazy({ route }) : lazy}
                       lazyPreloadDistance={lazyPreloadDistance}
                       navigationState={navigationState}
-                      style={sceneContainerStyle}
+                      containerStyle={sceneContainerStyle}
                     >
                       {({ loading }) =>
                         loading


daniel112:
Providing more context:
Value passed in on `sceneContainerStyle` is being overridden with absolute styles

example without `sceneContainerStyle`:
```js
        <TabView
          navigationState={{index, routes}}
          renderScene={renderScene}
          onIndexChange={setIndex}
        />
 // style output on SceneView: {"style": [undefined, {"bottom": 0, "height": undefined, "left": 0, "position": "absolute", "right": 0, "top": 0, "width": undefined}]}
```

examplewith `sceneContainerStyle`:
```js
        <TabView
          sceneContainerStyle={{
            width: 500,
            height: 500,
          }}
          navigationState={{index, routes}}
          renderScene={renderScene}
          onIndexChange={setIndex}
        />
 // style output on SceneView: {"style": [{"height": 500, "width": 500}, {"bottom": 0, "height": undefined, "left": 0, "position": "absolute", "right": 0, "top": 0, "width": undefined}]}
```

Not sure where that default style value is coming from tbh

src/SceneView.tsx
```diff
@@ -110,7 +110,7 @@ export default class SceneView<T extends Route> extends React.Component<
             : focused
             ? StyleSheet.absoluteFill
             : null,
-          style,
+          containerStyle,
```
daniel112, line 113: I ran into an edge case with this if you want to add a tab view that's not a full screen dimension.
if you add back the `styles` prop and add this line `!focused && style` right above `containerStyle`, it will allow the active scene content to dictate the height (unless you've set height on `TabView` styles)

```suggestion
          !focused && defaultStyle,
          containerStyle,
```


https://user-images.githubusercontent.com/3317835/124084979-67bb8900-da04-11eb-842b-98980a40a798.mp4

daniel112: Just so you know, without the suggested change. the heights will be messed up if you have multiple scenes

MonsterAnan:
I try you project,TabView in SectionList Component, isn't  fix my problem.

daniel112:
> I try you project,TabView in SectionList Component, isn't  fix my problem.
> 
> 

Try this branch https://github.com/knockaway/react-native-tab-view/tree/forked-3.0.2

vinceprofeta:
Is there an issue with @daniel112 solution? Can we get this merged?

satya164:
How does renaming the prop from `style` to `sceneContainerStyle` address the issue?

daniel112:
https://github.com/satya164/react-native-tab-view/pull/1198#issuecomment-870843828

@satya164

satya164:
@daniel112 that doesn't explain why changing the prop name is necessary.

satya164:
anyway, this PR doesn't explain any of the changes and doesn't provide any demos of what it fixes, so I'm leaning towards closing it. if someone can open a new PR providing a better explanation and demos etc. then I'll review that.

daniel112:
@satya164 because passing custom styles into "sceneContainerStyles" cannot actually override the position or size (height/width) styles

I agree this PR is still missing some changes. 
I can submit a PR of my branch from above with examples.
some people seem to have found it useful

satya164:
> because passing custom styles into "sceneContainerStyles" cannot actually override the position or size (height/width) styles

Still don't understand how just changing the name from `style` to `containerStyle` changes any behavior

bhattanmol101:
Really waiting for the fix asap can update react-native due to this need version 0.66.0

src/SceneView.tsx
raymondjacobson: This seems to cause performance degradation/ getting stuck halfway between tabs. Does anyone have thoughts?

j-cheung:
I understand! 

Changing the prop name from style to containerStyle fixes the issue because, for some reason, the style was passing in the properties of StyleSheet.absoluteFill.

```
diff --git a/node_modules/react-native-tab-view/src/SceneView.tsx b/node_modules/react-native-tab-view/src/SceneView.tsx
index 4648bf7..20ab78a 100644
--- a/node_modules/react-native-tab-view/src/SceneView.tsx
+++ b/node_modules/react-native-tab-view/src/SceneView.tsx
@@ -14,7 +14,7 @@ type Props<T extends Route> = SceneRendererProps &
     lazyPreloadDistance: number;
     index: number;
     children: (props: { loading: boolean }) => React.ReactNode;
-    style?: StyleProp<ViewStyle>;
+    sceneContainerStyle?: StyleProp<ViewStyle>;
   };
 
 type State = {
@@ -102,7 +102,7 @@ export default class SceneView<T extends Route> extends React.Component<
   };
 
   render() {
-    const { navigationState, index, layout, style } = this.props;
+    const { navigationState, index, layout, sceneContainerStyle } = this.props;
     const { loading } = this.state;
 
     const focused = navigationState.index === index;
@@ -120,7 +120,7 @@ export default class SceneView<T extends Route> extends React.Component<
             : focused
             ? StyleSheet.absoluteFill
             : null,
-          style,
+            sceneContainerStyle ? sceneContainerStyle : StyleSheet.absoluteFill,
         ]}
       >
         {
```

i have applied the patch above to workaround it for now 

but doesn't take into account [https://github.com/satya164/react-native-tab-view/pull/1198#discussion_r662052975](url)

bhattanmol101:
Changing the name fixes the tab height issue not sure why but it does work.

ildfreelancer:
@satya164 I think the root of this issue because
this line: https://github.com/satya164/react-native-tab-view/blob/629160ab86384f9fecc820e10d57fad0232b4a87/src/PagerViewAdapter.tsx#L15

For ios:
This is created Animated Component of Pager from Animated.createAnimatedComponent 

https://github.com/satya164/react-native-tab-view/blob/629160ab86384f9fecc820e10d57fad0232b4a87/src/PagerViewAdapter.tsx#L15

```
const AnimatedViewPager = Animated.createAnimatedComponent(ViewPager);
```

and take a look closer at version 5.4.24 which this package is using
https://github.com/callstack/react-native-pager-view/blob/931b19356b1b50183774b4c37610764615b8946a/src/utils.tsx

```
export const childrenWithOverriddenStyle = (children?: ReactNode) => {
  // Override styles so that each page will fill the parent. Native component
  // will handle positioning of elements, so it's not important to offset
  // them correctly.
  return Children.map(children, (child) => {
    const { props } = child as ReactElement;
    const newProps = {
      ...props,
      style: [
        props.style,
        {
          position: 'absolute',
          left: 0,
          top: 0,
          right: 0,
          bottom: 0,
          width: undefined,
          height: undefined,
        },
      ],
      collapsable: false,
    };
    return React.cloneElement(child as ReactElement, newProps);
  });
};
```

You can see that they are overriding style props, and this style will be passed into children props by render function

Therefore, this PR changes **style** to **different name** will actually work.

Other than this fix, we can fix from Pager from react-native-pager.

PR #1210, opened 2021-07-22
Add layoutDirection to AnimatedViewPager

aleppos:
### Motivation

Now react-native-pager-view accept layoutDirection, and react-native-tab-view doesn't support RTL so we need to pass layoutDirection to support rtl when I18nManager.isRTL is true.


aleppos:
This PR will fix this https://github.com/satya164/react-native-tab-view/issues/1149

aleppos:
This has been fixed from react-native-pager-view

PR #1237, opened 2021-07-30
Allow children dictate height

daniel112:

**Motivation**

Explain the **motivation** for making this change. What existing problem does the pull request solve?

**Test plan**

Demonstrate the code is solid. Example: the exact commands you ran and their output, screenshots / videos if the pull request changes UI.

Make sure you test on both platforms if your change affects both platforms.

The code must pass tests.

**Code formatting**

Look around. Match the style of the rest of the codebase. Run `yarn lint --fix` before committing.


daniel112:
@satya164 - opening PR with examples as requested - https://github.com/satya164/react-native-tab-view/pull/1198

src/SceneView.tsx
```diff
@@ -102,7 +103,13 @@ export default class SceneView<T extends Route> extends React.Component<
   };
 
   render() {
-    const { navigationState, index, layout, style } = this.props;
+    const {
+      navigationState,
+      index,
+      layout,
+      containerStyle,
+      style: defaultStyle,
```
satya164, line 111: Where is this `style` prop coming from? It was removed in only place which passed it in the PR

src/TabView.tsx
```diff
@@ -112,7 +112,7 @@ export default function TabView<T extends Route>({
                       lazy={typeof lazy === 'function' ? lazy({ route }) : lazy}
                       lazyPreloadDistance={lazyPreloadDistance}
                       navigationState={navigationState}
-                      style={sceneContainerStyle}
+                      containerStyle={sceneContainerStyle}
```
satya164, line 115: Why rename the prop?

src/SceneView.tsx
```diff
@@ -120,7 +127,8 @@ export default class SceneView<T extends Route> extends React.Component<
             : focused
             ? StyleSheet.absoluteFill
             : null,
-          style,
+          !focused && defaultStyle,
```
satya164, line 130: Not sure what this does given that nobody is passing the `style` prop and `containerStyle` does the same thing that `style` did but with a different name.

satya164 (changes requested):

src/SceneView.tsx
daniel112: After digging around, it comes from the class component itself. Regardless if we pass in our own prop called `style` into this component, it will have AT LEAST these styles (before my changes)
```
{"style": [undefined, {"bottom": 0, "height": undefined, "left": 0, "position": "absolute", "right": 0, "top": 0, "width": undefined}]}
```

where it says `undefined` as the first array item, that's where our passed in `style` prop would be

And the styles above gets merged into whatever we pass in to `style` prop, any style property that is `bottom/height/left/position/right/top/width` will always get overridden by the 
`{"bottom": 0, "height": undefined, "left": 0, "position": "absolute", "right": 0, "top": 0, "width": undefined}]}` styles

src/TabView.tsx
daniel112: because some style property will be overridden otherwise. See previous comment

src/SceneView.tsx
daniel112: i'm honestly not sure how and where the `style` prop is getting its data from, but it does, so I was working around it.

src/SceneView.tsx
yfunk: > 
> 
> Where is this `style` prop coming from? It was removed in only place which passed it in the PR

I don't completely understand the (quite complicated) structure of the `TabView` component, but as far as I can tell its coming from `Pager` aka. [`PagerViewAdapter` here](https://github.com/satya164/react-native-tab-view/blob/97c0857b534d08c12c06cb0c1b348a16c3ac8da9/src/PagerViewAdapter.tsx#L126), which wraps `SceneView`.

`sceneContainerStyle` is passed as the `style` prop to `SceneView` but somehow overwritten by the `style` prop from `PagerViewAdapter`. This causes `sceneContainerStyle` to not be applied at all, renaming the `style` prop from `SceneView` to anything else (could be `containerStyle` like here or even just `style2`) fixes that.

bayerlse:
Hey! When will this be available? :)

SpeedOfSpin:
I had this issue which pretty much stopped me from being able to use the component. I downloaded the source from the fork and it works great for me.

blackSooooo:
HI. I'm in trouble to solve this issue. When will it be applied?

demchenkoalex:
Don't think this fixes the issue entirely, it is working indeed, but there is this unwanted behaviour visible on an attached original video, where the height is being set only after swipe is done, and if heights of the content are different it does not look good (even more if you do swipe with a finger and you don't see any content until you release the finger). In v2 this was not the case, heights were set before swipe is done, so you could go back and forth between tabs with different heights without any visual glitches.

| Progress             |  Result |
:-------------------------:|:-------------------------:
Screenshot 2022-01-07 at 14 16 29|Screenshot 2022-01-07 at 14 16 42

fknop:
Also stuck without this fix, cannot upgrade to react-navigation 6, what is blocking?

pistou:
It's such a shame this fix is not merged yet. More than 6 months after.

src/SceneView.tsx
janicduplessis: The style is prop is modified here https://github.com/callstack/react-native-pager-view/blob/master/src/utils.tsx#L17 by react-native-pager-view.

beerent:
please merge this! I lost days working around this issue. This branch fixed my issue with zero code changes on my end!

bledar:
This PR is fixing the problem with the height of screens. I am using the fork version at the moment and hopefully this will be merged soon.

okwasniewski:
Hey @daniel112 can you sync this with the main branch? So we can get back to working on this and merge it.

daniel112:
> Hey @daniel112 can you sync this with the main branch? So we can get back to working on this and merge it.

@okwasniewski 
I no longer have access to the forked repo. 
So I forked it on my personal account and have an identical updated PR here:
https://github.com/satya164/react-native-tab-view/pull/1383

feel free to close this one when you're ready

PR #1245, opened 2021-08-16
fix: add pager initial index ref

pulecka:
Pager `initialPage` prop was set to the current index with every render.
Wrapped initial index in `Ref` so it stays unchanged during the whole component lifecycle.

**Motivation**

Android version of `react-native-pager-view` recently changed (callstack/react-native-pager-view#407) handling of the `initialPage` prop.
Previous versions handled initial page on the JS-side during `didMount` lifecycle method so it only ran once.
Now it is handled as native prop and selecting page using tab bar triggers 2 native current page handlers with one of them being not animated. Page transition is not animated at all as a result.

### DIFF
--- a/src/PagerViewAdapter.tsx
+++ b/src/PagerViewAdapter.tsx
@@ -49,6 +49,7 @@ export default function PagerViewAdapter<T extends Route>({
 
   const pagerRef = React.useRef<ViewPager>();
   const indexRef = React.useRef<number>(index);
+  const initialIndexRef = React.useRef<number>(index);
   const navigationStateRef = React.useRef(navigationState);
 
   const position = useAnimatedValue(index);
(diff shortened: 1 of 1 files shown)


MaximeCrp:
So with fixing this we would be able to use pager-view's initialPage prop? How? With setting the initial index in the navigationState prop of TabView?

src/PagerViewAdapter.tsx
```diff
@@ -124,7 +125,7 @@ export default function PagerViewAdapter<T extends Route>({
         {...rest}
         ref={pagerRef}
         style={[styles.container, style]}
-        initialPage={index}
+        initialPage={initialIndexRef.current}
```
rocwind, line 128: If the `index` get changed before `<ViewPager>` mounted: since only the very initial index being kept and passed to pager now, and the followed `index` change update will be missed ( `pagerRef.current?.setPage(index)` in `useEffect()` does nothing due to `pagerRef` not received yet), it may leave the selected page in a wrong state.

troZee:
I think, we can close this PR in favor of https://github.com/callstack/react-native-pager-view/pull/457

troZee:
https://github.com/callstack/react-native-pager-view/releases/tag/v5.4.5

PR #1252, opened 2021-08-23
fix: fix pager style

olvipi:
This PR corrects the style of the "TabView" root component when used with the "react-native-web" package in web applications.

code example
    <TabView
      style={{ height: 250 }}
      {...tabViewProps}
    />

before
image

after
image
image
image

### DIFF
--- a/src/TabView.tsx
+++ b/src/TabView.tsx
@@ -141,7 +141,8 @@ export default function TabView<T extends Route>({
 
 const styles = StyleSheet.create({
   pager: {
-    flex: 1,
+    flexGrow: 1,
+    flexShrink: 1,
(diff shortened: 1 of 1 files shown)


src/TabView.tsx
```diff
@@ -141,7 +141,8 @@ export default function TabView<T extends Route>({
 
 const styles = StyleSheet.create({
   pager: {
-    flex: 1,
+    flexGrow: 1,
+    flexShrink: 1,
```
okwasniewski, line 145: If your PR targets only web you should check if we are currently on web. Also did you test this on iOS and Android?

okwasniewski (changes requested):

PR #1260, opened 2021-09-10
Expose the style prop accepted by the Pager component

simonbuerger:
Pager component already has a style prop, this just exposes it to the end user

**Motivation**

I have a use case where part of the custom tab bar needed to overlap part of the scene. The only way to do it would be to add a negative top margin to the Pager view, however the style prop that is exposed on the Pager view is not exposed on the TabView so currently I can't achieve this use case.


okwasniewski (approved):
Thanks for the PR, looks good 👍🏻

PR #1276, opened 2021-10-09
(iOS) overdrag boolean prop for TabView

sungsong88:
**Motivation**

When swipeEnabled is true, sometimes on **iOS** the app does or doesn't want to allow for **overscrolling** after reaching **the very end or very beginning of pages**. `overdrag` is one of the props from **react-native-pager-view** for this, but it wasn't presented in **react-native-tab-view**. This is also important for the latest **@react-navigation/material-top-tabs** which uses **react-native-tab-view**.

**Test plan**
```JSX
import { SceneRendererProps, TabView } from 'react-native-tab-view';

export default function MaterialTopTabView({
  overdrag,
  ...
}: Props) {
  return (
    <TabView<Route<string>>
      overdrag={overdrag} // boolean: true or false. Optional. Defaults to false.
      ...
    />
  );
}
```
Before when it was not configurable, it's always over-scrollable.

https://user-images.githubusercontent.com/4941095/136645109-741c7352-d6ae-48fa-ba82-869b959826da.MP4

But now when it's set to false, over-scrolling can be prevented.

https://user-images.githubusercontent.com/4941095/136645121-b60a421a-eb7d-4e66-b8d2-b45852d844be.MP4

### DIFF
--- a/README.md
+++ b/README.md
@@ -335,6 +335,10 @@ Style to apply to the view wrapping each screen. You can pass this to override s
 
 Style to apply to the tab view container.
 
+##### `overdrag`
+
+Allows for over-scrolling after reaching the very end or very beginning of pages. It defaults to false (iOS only).
+
 ### `TabBar`
 
 Material design themed tab bar. To customize the tab bar, you'd need to use the `renderTabBar` prop of `TabView` to render the `TabBar` and pass additional props.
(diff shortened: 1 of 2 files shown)


Aryk:
On that note...there are several other properties that show up in typescript that are not transferred to the SceneView.

Something is definitely wrong here... we probably should be using the spread `...props` and passing it into the PagerComponent...

https://github.com/satya164/react-native-tab-view/blob/main/src/TabView.tsx#L79

okwasniewski:
Closing this since `overdrag` now defaults to `false` in react-native-pager-view. Passing more props to `PagerView` can be addressed in separate PR.

PR #1331, opened 2022-03-27
fix Vertical Scrollbar show on Tabbar

bdtren:
- Replicated in iOS. When I try scrolling on the tabBar there will be a vertical indicator show on the right side of it.
- This issue happened when I custom the TabBarItem.

Screen Shot 2022-03-27 at 22 02 44


okwasniewski (approved):
Looks good 👍🏻

PR #1350, opened 2022-05-20
feat: add index to 'onSwipeEnd' call

tironiigor:
**Motivation**

Having the index passed to the `onSwipeEnd` call allows us to use the latest index after the swipe event. This is useful if you keep the index state in your application and want to avoid rendering issues (animation and wrong tab selection) if you update it using the `onIndexChange` callback

**Test plan**

Add the `onSwipeEnd` call and make sure it's showing the correct value. It should be called after `onIndexChange` and it should have the same index number

### DIFF
--- a/src/PagerViewAdapter.tsx
+++ b/src/PagerViewAdapter.tsx
@@ -83,7 +83,7 @@ export default function PagerViewAdapter<T extends Route>({
 
     switch (pageScrollState) {
       case 'idle':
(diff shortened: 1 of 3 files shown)


okwasniewski:
Hey @tironiigor, Thanks for your PR unfortunately at the moment we can't merge your PR. It's inconsistent with the API and you can achieve same results with `onIndexChange` prop.

PR #1361, opened 2022-06-04
fix: next.js init error.

alantoa:

**Motivation** 

get this error info when use `React Native` + `Next.js`: 
image

Explain the **motivation** for making this change. What existing problem does the pull request solve?

- Fix the case where the width is 0 when `next.js` is initialized.

**Test plan**
- Check if the TabView load.


alantoa:
hi, @satya164!
thank you so much for this library!
we got this error using React Native + Next.js, hope you can review it, let's make your library even better!
( I've added a lot of features to this lib, which will be released soon!)

PR #1388, opened 2022-09-05
feat: add animationEnabled to TabView

szymonrybczak:
**Motivation**

According to this issue https://github.com/satya164/react-native-tab-view/issues/1371 I added the `disableChangeTabAnimation` to disable animation when changing tab.


okwasniewski (changes requested):

src/PagerViewAdapter.tsx
```diff
@@ -72,8 +79,14 @@ export default function PagerViewAdapter<T extends Route>({
...
     if (indexRef.current !== index) {
-      pagerRef.current?.setPage(index);
+      if (animationEnabled) {
+        pagerRef.current?.setPage(index);
+      } else {
+        pagerRef.current?.setPageWithoutAnimation(index);
+        position.setValue(index);
+      }
     }
+    // eslint-disable-next-line react-hooks/exhaustive-deps
```
satya164, line 89: Don't disable the lint rule. The lint rule is valid - the code in this function will break if `animationEnabled` isn't in the dependency array as it'll only respect what was set initially.

You can instead use `use-latest-callback` package for the `useCallback`, and include the dependency in the dep array for the `useEffect`.

satya164 (changes requested):

szymonrybczak:
@satya164 I updated my code as you said, could you please look at it again?

example/package.json
```diff
@@ -22,7 +22,8 @@
     "react-native": "0.69.5",
     "react-native-pager-view": "5.4.25",
     "react-native-safe-area-context": "4.3.1",
-    "react-native-web": "~0.18.7"
+    "react-native-web": "~0.18.7",
+    "use-latest-callback": "^0.1.5"
```
okwasniewski, line 26: This package is not needed inside `example`.

satya164 (changes requested):
This also needs to be handled in `PanResponderAdapter` for Web, where it should call `panX.setValue` instead of performing an animation

src/PagerViewAdapter.tsx
satya164: This comment should be removed

okwasniewski (approved):

src/PanResponderAdapter.tsx
```diff
@@ -81,22 +82,27 @@ export default function PanResponderAdapter<T extends Route>({
...
+            onIndexChangeRef.current(index);
+            pendingIndexRef.current = undefined;
+          }
+        });
+      } else {
+        panX.setValue(offset);
+        onIndexChangeRef.current(index);
+      }
 
       pendingIndexRef.current = index;
```
satya164, line 103: This should only be set when there is animation, so move it into the `if` block

src/PanResponderAdapter.tsx
```diff
@@ -81,22 +82,27 @@ export default function PanResponderAdapter<T extends Route>({
...
-          toValue: offset,
-          useNativeDriver: false,
-        }),
-      ]).start(({ finished }) => {
-        if (finished) {
-          onIndexChangeRef.current(index);
-          pendingIndexRef.current = undefined;
-        }
-      });
+      if (animationEnabled) {
```
satya164, line 85: This breaks the swipe gesture. The animation should only be disabled when not swiping similar to how it works on Android & iOS

src/types.tsx
```diff
@@ -55,4 +55,5 @@ export type PagerProps = Omit<
   swipeEnabled?: boolean;
   onSwipeStart?: () => void;
   onSwipeEnd?: () => void;
+  animationEnabled?: boolean;
```
satya164, line 58: Move the type next to `swipeEnabled`

README.md
```diff
@@ -339,6 +339,10 @@ Style to apply to the pager view wrapping all the scenes.
 
 Style to apply to the tab view container.
 
+### `animationEnabled`
```
satya164, line 342: The heading level isn't right, move this to next to `swipeEnabled` and adjust the heading level to match it (`#####`)

satya164 (changes requested):

src/PanResponderAdapter.tsx
szymonrybczak: Yes, you're right, I didn't noticed that, fixed ✅

mlecoq:
Hi, I have made a separated PR: see #1391 because I need to be able to deactivate animation only for distant tabs (I need to keep them for neighbor tabs)

satya164 (approved):

PR #1402, opened 2022-11-09
Fix: the tab header not sync with scene swipe when header is long

gongdao123:

**Motivation**

- The Tab Header is utilizing FlatList for rendering propose.
- FlatList will NOT try to render ALL of the content if the content is long for performant reasons.
- With the previous code, the tab header moves ONLY when all of the header content is rendered.
**These cause the header will not sync with the scene swipe when the header is long**

output

**Test plan**

After the patch, the tab header can follow the screen swipe, with no performance drawback.
output1

**Code formatting**

Look around. Match the style of the rest of the codebase. Run `yarn lint --fix` before committing.

### DIFF
--- a/src/TabBar.tsx
+++ b/src/TabBar.tsx
@@ -62,6 +62,7 @@ export type Props<T extends Route> = SceneRendererProps & {
   contentContainerStyle?: StyleProp<ViewStyle>;
   style?: StyleProp<ViewStyle>;
   gap?: number;
+  testID?: string;
 };
 
 type FlattenedTabWidth = string | number | undefined;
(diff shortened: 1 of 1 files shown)


okwasniewski:
Hey, thanks for your PR. Unfortunately I don't think that using `setTimeout` won't have any performance drawback. Therefore I think that this should be solved differently. Can you open an issue for this I will take a look into this.

gongdao123:
> Hey, thanks for your PR. Unfortunately I don't think that using `setTimeout` won't have any performance drawback. Therefore I think that this should be solved differently. Can you open an issue for this I will take a look into this.

Thx for your reply, I've created issue #1405.

Just in case. Want to clarify here that the `clearTimeout` and `setTimeout` are very basic debounce mechanisms to make sure the `setTabWidths` will ONLY trigger once in one batch, which should not have some performance drawback?

okwasniewski:
Closing this PR, since I've implemented a proper fix for this here: #1409

